PR #4, opened 2020-11-26
Add documentation and setup readthedocs hosting

AlexIoannides:
Which will be hosted [here](https://bodywork.readthedocs.io/en/latest/).


AlexIoannides:
Not going to ask for an explicit review as you've already reviewed the docs and the rest is all config for readthedocs (which hopefully I've got right...).

PR #25, opened 2021-02-17
Add high-level interface info to Overview section of docs

AlexIoannides:
[Bodywork - Overview.pdf](https://github.com/bodywork-ml/bodywork-core/files/5996984/Bodywork.-.Overview.pdf)


Marios85 (commented):
The new section is perfect!  I have a suggestion for the diagram, I think that the Bodywork logo should be above the arrows and the .py boxes should be surrounded by a box with the k8s logo. I say this because I think it makes it clearer that Bodywork is doing the lifting from Git and deploying to the k8s cluster. 

On the flip-side, my suggestion might make it look like Bodywork doesn't have control after it is deployed. 

What do you think?

Marios85 (approved):

PR #26, opened 2021-02-18
Setup docs for use with Material theme for MkDocs

AlexIoannides:
- change MkDocs config
- change layout/structure
- modify images

The [Materials](https://squidfunk.github.io/mkdocs-material/) theme enables new styling options that I have used selectively to make the docs easier to read and use.

Check the new version [here](https://bodywork.readthedocs.io/en/add-materials-theme-to-docs/) 🙂


Marios85 (approved):
😎

PR #49, opened 2021-03-03
Modify workflow with ingress logic

### LINKED ISSUE #35: Modify workflow runner to handle ingress resources during execution
**Story**
As a ML engineer, I would like service ingress to be setup (or updated) as required, so that I can expose my services to the world beyond the k8s cluster.

**Tasks**
- Modify `workflow.py` to create Ingress resources alongside Service resources and update changes accordingly.
- Check that the same can be achieved with Services (updates to things like ports might be lacking).
- Ensure that a `WARNING` is logged when ingress is configured, but there is no NGINX ingress controller deployed in the cluster.

AlexIoannides:
This PR introduces ingress-handling logic into `workflow.py` and extends the integration tests in `test_k8s_with_cluster.py` to cover the new logic.

It closes #35


AlexIoannides:
Note, that I have updated [bodywork-test-project](https://github.com/bodywork-ml/bodywork-test-project) so that `stage-3` has ingress and `stage-4` does not (they are both services) and exploited this in the integration tests.

Please ignore the failed CodeCov status. When I was sorting-out the GitHub nonsense I killed one of the pipelines for `v0.3.0` and the CodeCov service didn't receive the notification from the integration tests (it is supplied with two files to collate a holistic picture of our testing).

As it is, it's next to impossible to test a stage that is marked with `INGRESS=True` in an initial deployment, followed by a change to `INGRESS=False` on subsequent deployments. So, we loose ~0.2% coverage (two lines). Still > 99%...

Marios85 (approved):
All good!

Issue #40, opened 2021-02-26
Enable ad hoc remote workflow execution


AlexIoannides:
**Story**
As a ML engineer, I would like to be able to trigger workflows on an ad hoc basis, without having to have the workflow-controller run locally, so that my machines (e.g. CICD runners) do not have to dedicate non-cluster resources for running workflows.

**Tasks**
- Create a `bodywork.k8s.jobs.configure_workflow_job` function for defining a workflow execution job, using `bodywork.k8s.cronjobs.configure_cronjob` as a reference.
- Create a `bodywork.cli.deploy` module that contains a function for triggering workflow job.
- Create a `bodywork deploy` command in `bodywork.cli.cli` for providing this functionality from the CLI.

PR #55, opened 2021-03-06
Origin/37 improve exception handling and error messages for failed git clones

### LINKED ISSUE #37: Improve exception handling and error messages for failed git clones
**Story**
As a ML engineer, I would like to know when a workflow has failed because it is private and the SSH credentials are missing, so that I can take the necessary steps to correct the problem.

**Task**
Improve the exception handling in `git.py`, so that users are better informed.

Marios85:
This PR improves the exception handling in git.py, so that users receive the git error when trying to download a git repo. This closes #37 

Test fixture was also fixed to be compatible with Windows OS.


tests/unit_and_functional/conftest.py
```diff
@@ -75,3 +77,22 @@ def k8s_env_vars() -> Iterable[bool]:
...
+
+    Usage : ``shutil.rmtree(path, onerror=onerror)``
+    """
+    import stat
+    if not os.access(path, os.W_OK):
+        # Is the error an access error ?
+        os.chmod(path, stat.S_IWUSR)
+        func(path)
+    else:
+        raise
```
AlexIoannides, line 98: You need to specify what type of exception to raise - e.g. `raise RuntimeError('this did not work)`. 

Without this, you'll get an exception raised due to the bad exception definition - e.g. `RuntimeError: No active exception to reraise`

tests/unit_and_functional/conftest.py
```diff
@@ -55,15 +56,16 @@ def setup_bodywork_test_project(
     try:
         run(['git', 'init'], cwd=project_repo_location, check=True)
         run(['git', 'add', '-A'], cwd=project_repo_location, check=True)
-        run(['git', 'commit', '-m', '"test"'], cwd=project_repo_location, check=True)
-        run(['mkdir', bodywork_output_dir], check=True)
+        run(['git', 'commit', '-m', '"test"'], cwd=project_repo_location, check=True, capture_output=True)
```
AlexIoannides, line 59: Could you please add `encoding='utf-8` to the options please?

This will stop the output being written as bytes. You're basically correcting my original work here - sorry...

tests/unit_and_functional/conftest.py
```diff
@@ -75,3 +77,22 @@ def k8s_env_vars() -> Iterable[bool]:
...
+    Error handler for ``shutil.rmtree``.
+
+    If the error is due to an access error (read only file)
+    it attempts to add write permission and then retries.
+
+    If the error is for another reason it re-raises the error.
+
+    Usage : ``shutil.rmtree(path, onerror=onerror)``
+    """
+    import stat
```
AlexIoannides, line 92: It's considered bad practice to place imports anywhere other than at the top of a module file.

tests/unit_and_functional/conftest.py
```diff
@@ -75,3 +77,22 @@ def k8s_env_vars() -> Iterable[bool]:
     finally:
         yield True
     del os.environ['KUBERNETES_SERVICE_HOST']
+
+def onerror(func, path, exc_info):
```
AlexIoannides, line 81: `on_error` would be better, despite the naming convention used in the Python standard library (which is not always PEP8, due to it being written before PEP8...).

tests/unit_and_functional/conftest.py
```diff
@@ -75,3 +77,22 @@ def k8s_env_vars() -> Iterable[bool]:
...
+    """
+    Error handler for ``shutil.rmtree``.
```
AlexIoannides, line 83: The first line needs to go above and come directly after the `"""`

AlexIoannides (commented):
Okay, so I tried to test this using a private repo I created for the task (`bodywork-ml/37-test`),

```shell
bodywork workflow \
    --namespace=37-test \
    --bodywork-docker-image=bodyworkml/bodywork-core:0.3.0.alpha3-dev \
    https://github.com/bodywork-ml/37-test-project.git \
    master
```

Where bodywork hasn't been configured with any GitHub credentials.

It doesn't work, but this is not your code's fault 😞 

This all stems from Git requesting a username and password on the command-line, when a repo is private and there are no credentials it can use. Here are the logs from a remote workflow execution (which I executed using a cronjob as we can't do this any other way until #40 is completed...),

```text
git version 2.20.1
Cloning into 'bodywork_project'...
fatal: could not read Username for 'https://github.com': No such device or address
2021-03-08 11:36:09,315 - ERROR - workflow.run_workflow - failed to execute workflow for master branch of project reposotory at https://github.com/bodywork-ml/37-test: git clone failed - calling ['git', 'clone', '--branch', 'master', '--single-branch', 'https://github.com/bodywork-ml/37-test', PosixPath('bodywork_project')] returned None
```

My original thinking was that we could capture stderror and regex it for `fatal: could not read Username`, and then convert the error message into something meaningful for the user. However, I have spent the last 2h trying to debug and test this idea by launching containers on k8s... with no f*cking success.

My current thinking, is that we write a function along the lines of,

```python
def is_configured_for_private_github_repos() -> None:
    if not os.environ.get(BODYWORK_GITHUB_SSH_PRIVATE_KEY):
        log.warning('not configured for use with private GitHub repos')
```

And then call it before the `git clone` command. This way, users can at least easily figure-out why things are failing, because the available info is unintelligible at the moment.

It's deeply irritating how the smallest things can blow-up like this...

AlexIoannides:
image

Another problem that is part of this issue, is that when `download_project_code_from_repo` throws an exception it is never handled in any way, as `bodywork.cli.cli.py` doesn't do any logging and just return an exit code of 1 (by design). Hence, we probably need to get `download_project_code_from_repo` into the try-catch block and make sure that any exception that it throws are logged!

AlexIoannides:
To summarise the above, there are two additional tasks:

1. check for the presence of the `BODYWORK_GITHUB_SSH_PRIVATE_KEY` environment variable (this comes from `constants.py`) and log an info or warning message to let the user know if private repos will not work, etc.
2. modify `bodywork.stage.run_stage` so that `download_project_code_from_repo` is in a `try/except` block with any errors logged.

You may well have better ideas, but this is as far as I can get right now!

tests/unit_and_functional/conftest.py
Marios85: 👍

tests/unit_and_functional/conftest.py
Marios85: No worries, done!

tests/unit_and_functional/conftest.py
Marios85: I missed that, sorry!

tests/unit_and_functional/conftest.py
Marios85: 👍

Marios85:
I'm going to see if I can replicate the exact error (missing username) you are getting with this test repo. I have noticed with a specific 'error' whilst testing that sometimes it's written to stdout because Git doesn't view it as an error per se. Then I can try work something out. 

However I still agree that we should have a method in for checking BODYWORK_GITHUB_SSH_PRIVATE_KEY and I'll also add the try/except block in.

AlexIoannides:
Yes - when you test it locally and just hit 'enter' for the username and password requests, Git returns with an exit code of 0 and there's no error...

Marios85:
Previous suggestions have been implemented and i've tested the error message for the missing credentials (username) and it's captured in the output. 

I have modified the run_stage exception handling to catch all exceptions so that it is always logged and a bodyworkstagefailure exception is always raised.

src/bodywork/git.py
```diff
@@ -25,12 +25,13 @@
...
 
 from .constants import DEFAULT_PROJECT_DIR, SSH_DIR_NAME, SSH_GITHUB_KEY_ENV_VAR
+from .logs import bodywork_log_factory
 
 
 def download_project_code_from_repo(
-    url: str,
-    branch: str = 'master',
-    destination: Path = DEFAULT_PROJECT_DIR
+        url: str,
```
AlexIoannides, line 32: Don't need the tabs - see 3rd code block [here](https://black.readthedocs.io/en/stable/the_black_code_style.html#code-style).

It's a bit odd, but you get used to it.

src/bodywork/git.py
```diff
@@ -41,18 +42,23 @@ def download_project_code_from_repo(
...
-        setup_ssh_for_github()
-
+    try:
+        if (get_connection_protocol(url) is ConnectionPrototcol.SSH
+                and get_remote_repo_host(url) is GitRepoHost.GITHUB):
+            setup_ssh_for_github()
+        elif SSH_GITHUB_KEY_ENV_VAR not in os.environ:
+            log.warning('Not configured for use with private GitHub repos')
+    except Exception as e:
+        msg = f'git clone failed - Unable to setup SSH for Github: {e}'
```
AlexIoannides, line 57: Technically speaking, the `git clone` hasn't failed. Maybe,

> `Unable to setup SSH for GitHub and you are trying to connect via SSH`

?

tests/unit_and_functional/conftest.py
```diff
@@ -53,17 +55,18 @@ def setup_bodywork_test_project(
...
-    except CalledProcessError as e:
-        raise RuntimeError(f'Cannot create test project Git repo - {e}.')
-    yield True
-    # TEARDOWN
-    run(['rm', '-rf', '.git'], cwd=project_repo_location)
-    run(['rm', '-rf', cloned_project_repo_location])
-    run(['rm', '-rf', bodywork_output_dir])
+        run(['git', 'init'], cwd=project_repo_location, check=True, encoding='utf-8')
+        run(['git', 'add', '-A'], cwd=project_repo_location, check=True, encoding='utf-8')
+        run(['git', 'commit', '-m', '"test"'], cwd=project_repo_location, check=True, capture_output=True, encoding='utf-8')
```
AlexIoannides, line 60: This line is 117 character long - can you split it over two lines?

src/bodywork/stage.py
```diff
@@ -234,24 +234,24 @@ def run_stage(
...
+        run(['python', stage.executable_script_path], check=True, encoding='utf-8')
         log.info(f'successfully ran stage={stage_name} from {repo_branch} branch of repo'
                  f' at {repo_url}')
-    except CalledProcessError as e:
-        stage_failure_exception = BodyworkStageFailure(stage_name, e.cmd, e.stderr)
+    except Exception as e:
+        stage_failure_exception = BodyworkStageFailure(stage_name, e)
         log.error(stage_failure_exception)
-        raise stage_failure_exception from e
+        raise stage_failure_exception
```
AlexIoannides, line 248: Shouldn't we leave in the `from e` so that we preserve the trace-back?

tests/unit_and_functional/conftest.py
```diff
@@ -75,3 +78,19 @@ def k8s_env_vars() -> Iterable[bool]:
...
+
+def on_error(func, path, exc_info):
+    """Error handler for ``shutil.rmtree``.
+
+    If the error is due to an access error (read only file)
+    it attempts to add write permission and then retries.
+
+    If the error is for another reason it re-raises the error.
+
+    Usage : ``shutil.rmtree(path, onerror=onerror)``
```
AlexIoannides, line 90: Typo - should be `..., onerror=on_error)`

tests/unit_and_functional/test_git.py
```diff
@@ -46,11 +47,14 @@ def test_that_git_project_repo_can_be_cloned(
     except Exception:
         assert False
 
-
 def test_that_git_project_clone_raises_exceptions():
     with raises(RuntimeError, match='git clone failed'):
         download_project_code_from_repo('file:///bad_url')
 
+@patch('bodywork.git.setup_ssh_for_github')
+def test_that_git_project_clone_returns_git_error_in_exception(mock_setup_method: MagicMock):
```
AlexIoannides, line 55: The line is 94 characters long - could you split it over two lines?

My laptop screen is 13" - these things get amplified!

If you think there's a line that really should stay whole for readability, add `  #noqa` to it (with the leading 2 spaces). This is how you flag to flake8 that it can be ignored.

AlexIoannides (changes requested):
This is brilliant - a huge improvement. 

I have a few minor comments, because I am anally-retentive, but you don't need to seek any more approval from me - merge when you're ready.

src/bodywork/git.py
Marios85: I didn't touch this code, not sure why it's come up as changed. I have fixed it.

src/bodywork/git.py
Marios85: 👍

tests/unit_and_functional/conftest.py
Marios85: Well spotted!

src/bodywork/stage.py
Marios85: Sorry I misunderstood the Python docs when it said: 
'Exception chaining happens automatically when an exception is raised inside an except or finally section'
That's now been fixed.

tests/unit_and_functional/conftest.py
Marios85: 👍

tests/unit_and_functional/test_git.py
Marios85: My laptop is only 14" and that line finishes close to the middle and that's in Pycharm with the side panel displayed.

What have you got line length set to? Black recommends '90ish'.

Marios85:
Thanks! I will add a flake8 linter to my IDE and see if I can get one for Black so I can automatically format the files and not drive you crazy having to correct it 🙂

tests/unit_and_functional/test_git.py
AlexIoannides: 90 👍

Sorry, I meant side-by-side when doing a diff.

src/bodywork/git.py
AlexIoannides: Odd

AlexIoannides (approved):

PR #61, opened 2021-03-09
Bug fixes - correct argument passing in stage config exceptions

### LINKED ISSUE #47: Arguments to BodyworkStageConfig exceptions are not passed correctly in stage.py
Leading to log messages that make no sense.

For example, in line 161 of `stage.py` we have:

```python
time_param_error = BodyworkStageConfigError(
    'MAX_STARTUP_TIME_SECONDS',
    'batch',
    name
)
```

Which should actually be,

```python
time_param_error = BodyworkStageConfigError(
     name,
     'service'
    'MAX_STARTUP_TIME_SECONDS'
)
```

Check all uses of `BodyworkStageConfigError` in `stage.py` and fix where necessary.

AlexIoannides:
This PR correct the errors configuring the `BodyworkStageConfig` exceptions in `stage.py`. It also modifies the associated unit tests, so that this cannot happen again and in doing so closes #47


Marios85 (approved):
👍

PR #67, opened 2021-03-25
Refactor all Bodywork configuration

AlexIoannides:
This PR contains a single, but very large unit of refactoring. It removes/relocates the logic contained in:

- `bodywork.workflow.BodyworkProject`
- `bodywork.stages.Stage`
- `bodywork.stages.BatchStage`
- `bodywork.stages.ServiceStage`

And relocates it into complete refactoring of `bodywork.config.BodyworkConfig`.  This scope of this class has changed enormously - because, it now opens and parses a YAML configuration file containing **all** of the configuration, for the entire project. 

The change to a single config file enabled the opportunity to validate the config for the entire project in one pass, before any workflow has been started - this is now what is happening in `bodywork.config.BodyworkConfig` and why it is so large. 

The Builder pattern has been implemented in `bodywork.config.BodyworkConfig` to create an easily extensible Bodywork config validator. Collectively, this refactoring has brought the following major advantages and new features:

- We no longer need `config.ini` and `requirements.txt` files in every stage directory. This completely separates code from config as can be seen in the new [bodywork-test-project](https://github.com/bodywork-ml/bodywork-test-project), which is now structured as a conventional Python project with a `bodywork.yaml` in the root directory - in precisely the same manner as many CICD services. that hook into GitHub, require.
- We now have the option of breaking the relationship between the name given to a stage, and the name of the directory that the executable module resides in. For any given stage, you now just provide the path to the executable to be used for that stage. This will allow users to re-use an executable module for multiple stages, without having to copy-paste anything. Also, this way Bodywork imposes absolutely **nothing** on the structure of the codebase 🙂 To build on this flexibility, I built-in the capability to pass parameters to each executable module, via the config file, so that a single executable module can be re-used in different ways, depending in which stage it's running in. This is demonstrated in [bodywork-test-project](https://github.com/bodywork-ml/bodywork-test-project), for `stage_1`.
- Because all the config can be validated in one pass, it has been possible to accumulate a running log of all config errors. This has been used to implement a new CLI command - `$ bodywork validate` that prints a validation report to stdout, that can be used to debug the config, before deployment is tested. This is great for user-experience!
- It is now much, much easier to extend and test config related code, without having to spawn new test resources (e.g. directories containing `config.ini` and `requirements.txt` files), each time a new parameter is added. This is basically how I slipped-in new features like passing `args` in config, so quickly.
- All the config and validation is now in one module, under a single interface, which makes the codebase easier to navigate and comprehend.


src/bodywork/cli/cli.py
```diff
@@ -545,3 +567,31 @@ def setup_namespace(args: Namespace) -> None:
...
+        sys.exit(0)
+    except FileExistsError as e:
+        print(f'--> {e}')
+        sys.exit(1)
+    except BodyworkConfigParsingError as e:
+        print(f'--> {e}')
+        sys.exit(1)
+    except BodyworkConfigMissingSectionError as e:
+        print(f'--> {e}')
+        sys.exit(1)
```
Marios85, line 592: If these three exceptions are going to be treated the same, then just put them in one except block:

except ex1, ex2, ex3

Marios85 (changes requested):
This is a lot nicer having all the config in one file and it makes the testing cleaner too 🙂 

I think the Stage objects should be in stage.py and not in config.py. Config should be just for obtaining the config from file and validating that config. You would naturally expect the Stage objects to be in the file with the same name as well. 

Otherwise everything is looking "bang on" as per usual.

I also have a few suggestions: 

1. If I want to use the same resources for each stage, I will have to write that same information again for each stage. It would be nice if I could enter those details once in the top level and then it's automatically applied to each stage. Then maybe I can override those settings for each stage if I choose to enter them. This would reduce repetition and the size of the config massively if I had many stages. I would say this is a 'Nice To Have'

2. Instead of having the custom validator that you have written, can we not validate the YAML against a schema? I've seen some mention of [Cerberus ](https://docs.python-cerberus.org/en/stable/schemas.html)or even JSON validators that work with YAML to achieve this. This would be remove a huge chunk of code and any new config parameters added will just be adding to the schema file as opposed to writing code for it.

src/bodywork/cli/cli.py
AlexIoannides: Yes 🙂

AlexIoannides:
I've spent ~5h implementing Cerberus. At face-value it looked precisely like what we need, but I've now decided to give - here is my reasoning:

- We have a very complex schema - defining it has needed a Python dictionary that is  ~200 lines long!
- Parsing the errors is a f*cking nightmare - we're looking at a non-trivial amount of code (and time) required to get them into a human-readable format, which is probably quite brittle - we're one library update away from needing to re-write it.
- Because there are some complex dependencies in our schema (that can't be easily modelled with Cerberus), it has to be re-jigged to work. This results in YAML that is even more verbose, whereas we can use some nice shorthand tricks when we're rolling with our own code.

So, I vote for keeping it as-is for now and maybe re-factoring it later if a better solution comes along.

AlexIoannides:
On point (1) - this is a nice idea, but given the additional complexity involved to implement this 'power-user' feature, I say we leave it for now. What concerns me more, is that explaining/communicating the config gets harder. We ought to keep things simple, for now - the YAML is intimidating enough as it is!

AlexIoannides:
On the location of `Stage` and the classes that inherit from it - I feel quite strongly that these should remain in `config.py`, as they are highly related - logically - with `BodyworkConfig`. For example, they are explicitly designed to expose `missing_or_invalid_param` to `BodyworkConfig`. And they do nothing more than validate config and expose it via an API, just like the other classes in `config.py`. No unvalidated config should leave this module, etc.

That said, several names have now become misleading. So, I've done the following refactoring of names, to try and address this problem:

Within `config.py`:
- `Project` -> `ProjectConfig`
- `Logging` -> `LoggingConfig`
- `Stage` -> `StageConfig`
- `BatchStage` -> `BatchStageConfig`
- `ServiceStage` -> `ServiceStageConfig`

Of Python modules within the `bodywork` folder:
- `stage.py` -> `stage_execution.py`
- `workflow.py` -> `workflow_execution.py`

I think (hope) this is a lot clearer and more representative of what is going on, where. What do you think?

Marios85 (approved):
That sounds like Cerberus doesn't do what it says on the tin. A schema validator should not force you to change your config at all so that's a big 'No' especially as I think our YAML is quite simple. 

I'm happy to write up a Technical Debt ticket for this and tackle it later. I am surprised that there isn't a proper library yet to validate YAML and this actually makes XML look good. [RX ](http://rx.codesimply.com/) is the other option and seems to have versions in multiple languages so maybe we can try this out when we have the time.

As for point 1, that's cool, it's a "Nice to have".  I'll write up the ticket for it.

Regarding the Stage object, the renaming works and makes things a lot more readable and intuitive.

Cheers for making the changes! 😎

AlexIoannides:
Cerberus is very good and I'd be more than happy to use it. Now that I think about it, I could have got more out of it by breaking the schema down into 'sub-schema' to make validating the nested objects easier - trying to do it in one pass was asking too much.

But I don't think the codebase would get much simpler - were essentially outsourcing unit tests for validation to Cerberus.

The big advantage we have in the roll-your-own approach, is that we can easily tag errors to a global lost, from any part of the validation. Cerberus's output is great, but it's heavily nested. To make Cerberus work elegantly, I think you need to avoid validating nested objects from the top-down.

Marios85:
It should make the codebase simpler, instead of all that validate code it should be a one liner:  _Validate(config file, Schema)_.  Maybe some code to format the output but all it needs to show is that there are errors on this line/field to the user. 

I don't know how Cerberus works but this is what I would expect from a half decent schema validator (at least this is how it works with XML). 

I've created the Tech Debt ticket for this.

AlexIoannides:
Okay, I have implemented Cerberus in a way that works well for this project - I hope you agree!

Marios85:
This looks much nicer! Thanks for making the changes! 🙂

PR #68, opened 2021-03-29
Upgrade Python to v3.8 and upgrade k8s Python client

AlexIoannides:
I've structured this PR to be relative to the `yaml-config-refactor` branch, so that it's tested with the latest updates, whist being easy to see the diffs.


Marios85 (approved):
All good!

PR #79, opened 2021-04-24
#72 Gitlab integration

### LINKED ISSUE #72: Add support for Git repos hosted on GitLab
**Description**
_As a machine learning engineer, I would like to be able to use public and private repos hosted on GitLab with Bodywork, because I cannot use GitHub at my place of work and this prevents me from adopting Bodywork._

**Tasks**
- create a Bodywork GitLab account and create create a copy of the [bodywork-test-project](https://github.com/bodywork-ml/bodywork-test-project), or similar.
- extend the `bodywork.git` module to work with GitLab repos, either public or private (via [SSH](https://docs.gitlab.com/ee/ssh/)).
- ensure that a `SSH_GITLAB_KEY_ENV_VAR` will be injected as an environment variable in `bodywork.k8s.batch_jobs.configure_batch_stage` and `bodywork.k8s.service_deployments.configure_service_stage_deployment`, or consider refactoring `SSH_GITLAB_KEY_ENV_VAR` into `SSH_GIT_KEY_ENV_VAR`, so that it can be used with _any_ remote Git repository host.
- add an integration test that uses a GitLab repo.
- update docs accordingly.

Marios85:
PR to close #72 thereby adding Gitlab integration to Bodywork

Added support for Gitlab by enabling dynamic addition of host keys to the SSH known_hosts file. 

Before adding a new Git provider to the list of trusted hosts, their fingerprint is checked and cross referenced against a hard-coded copy of their fingerprint (publicly available on their websites) to ensure that the key comes from a genuine source.

Tests have been updated to work by reading in the SSH private key file from a user into the private key environment variable.

Also included in this PR:

  - Integration Test refactoring.
  - Amending the CircleCI config so that the test results are stored in a readable format.
  - Reformatting the whole code base with Black.


src/bodywork/git.py
```diff
@@ -44,102 +50,160 @@ def download_project_code_from_repo(
...
+    if hostname in fingerprints:
+        try:
+            server_key = run(
+                ["ssh-keyscan", "-t", "rsa", hostname],
+                check=True,
+                capture_output=True,
+                encoding="utf-8",
+            ).stdout
+            fingerprint = run(
+                ["ssh-keygen", "-l", "-f", "-"],
```
AlexIoannides, line 191: What's does the `"-"` do?

I couldn't work it out from reading the ssh-keygen man pages.

tests/unit_and_functional/conftest.py
```diff
@@ -61,11 +72,12 @@ def setup_bodywork_test_project(
             capture_output=True, encoding='utf-8')
         os.mkdir(bodywork_output_dir)
         yield True
-    except CalledProcessError as e:
+    except Exception as e:
         raise RuntimeError(f'Cannot create test project Git repo - {e.output}.')
     finally:
         # TEARDOWN
         shutil.rmtree('{}/.git'.format(project_repo_location), onerror=on_error)
+        shutil.rmtree('{}/.git'.format(cloned_project_repo_location), ignore_errors=True, onerror=on_error)
```
AlexIoannides, line 80: Could we stick to using f-strings - i.e. `f'{cloned_project_repo_location}/.git'`. I believe that the `.format` approach is slowly being phased-out. I failed to notice this in the above line when I reviewed it the first time around.

src/bodywork/git.py
Marios85: It won't be in there, it's required for piping in the input from the previous command using the input=server_key.

tests/unit_and_functional/conftest.py
Marios85: Yeah this shouldn't be different from the rest.

src/bodywork/git.py
AlexIoannides: Is that a generic Linux thing?

Is there any chance of making it more explicit, because I'll  never remember that.

VERSION
```diff
@@ -1 +1 @@
-2.0.2
+2.1.0
```
AlexIoannides, line 1: 2.1.0alpha1

src/bodywork/git.py
Marios85: I think it is and it works on Windows as well. I don't really know how to make it more explicit apart from in the comments. Would an explanation in the comment suffice?

VERSION
Marios85: Oops 😬

tests/unit_and_functional/conftest.py
```diff
@@ -81,6 +93,26 @@ def k8s_env_vars() -> Iterable[bool]:
...
+        if private_key.exists():
+            os.environ[SSH_PRIVATE_KEY_ENV_VAR] = private_key.read_text()
+        else:
+            raise RuntimeError('cannot locate private SSH key to use for GitHub')
+
+
+@fixture(scope='function')
+def set_gitlab_ssh_private_key_env_var() -> None:
+    if SSH_PRIVATE_KEY_ENV_VAR not in os.environ:
+        private_key = Path.home() / '.ssh/id_rsa_e28827a593edd69f1a58cf07a7755107'
```
AlexIoannides, line 109: Where does this funny name come from `id_rsa_e28827a593edd69f1a58cf07a7755107`?

tests/unit_and_functional/test_git.py
```diff
@@ -24,22 +24,22 @@
...
+    setup_ssh_for_git_host,
+    get_ssh_public_key_from_domain,
 )
 
 
 def test_that_git_project_repo_can_be_cloned(
     setup_bodywork_test_project: Iterable[bool],
     project_repo_connection_string: str,
-    cloned_project_repo_location: Path
+    cloned_project_repo_location: Path,
```
AlexIoannides, line 42: I'm not sure that comma is meant to be there.

tests/unit_and_functional/test_git.py
```diff
@@ -48,74 +48,113 @@ def test_that_git_project_repo_can_be_cloned(
...
+    setup_bodywork_test_project: Iterable[bool],
+    gitlab_repo_connection_string: str,
+    cloned_project_repo_location: Path,
+    set_gitlab_ssh_private_key_env_var: None,
+):
+    try:
+        ssh_dir = Path(".") / SSH_DIR_NAME
+        os.makedirs(ssh_dir, exist_ok=False)
+        with open(os.path.join(ssh_dir, "id_rsa"), "w"):
+            pass
```
AlexIoannides, line 77: `pass`?

Is this a WIP?

tests/unit_and_functional/conftest.py
Marios85: That is the filename that CircleCI gives the key (a hash of some sorts) because we now have multiple RSA keys. We have no control over this.

tests/unit_and_functional/test_git.py
Marios85: Good spot!

tests/unit_and_functional/test_git.py
Marios85: This is no longer needed, it was required to create a blank file in my previous implementation when the env variable for the private key was removed. 

Another good spot! 🙂

tests/unit_and_functional/conftest.py
AlexIoannides: Won't the tests fail locally then?

src/bodywork/git.py
AlexIoannides: A comment at the very least please.

I thought you could just pass is the filename from the previous command that it's implicitly referring to. Have I misunderstood what's going on?

.circleci/config.yml
```diff
@@ -28,36 +28,50 @@ jobs:
           command: | 
             git config --global user.name "Bodywork Machine Learning Ltd"
             git config --global user.email "info@bodyworkml.com"
+      - add_ssh_keys:
```
AlexIoannides, line 31: We don't need this now the tests are split?

.circleci/config.yml
```diff
@@ -28,36 +28,50 @@ jobs:
...
-          command: tox -e py38-k8s_integration_tests -- -s -vv
+          command: tox -e py38-integration_tests -- -s -vv
       - run:
           name: Code Coverage upload
           command: bash <(curl -s https://codecov.io/bash)
   build-dev-docker-image-push-to-dockerhub:
     machine: true
     steps:
       - checkout
+      - add_ssh_keys:
```
AlexIoannides, line 71: Should no longer be needed.

.circleci/config.yml
```diff
@@ -68,6 +82,10 @@ jobs:
...
```
AlexIoannides, line 85: Delete.

src/bodywork/git.py
```diff
@@ -44,102 +50,162 @@ def download_project_code_from_repo(
...
+                setup_ssh_for_git_host(hostname)
+            else:
+                raise ValueError(
+                    f"Unable to derive hostname from URL {url}. Please check "
+                )
+        elif SSH_PRIVATE_KEY_ENV_VAR not in os.environ:
+            log.warning("Not configured for use with private GitHub repos")
     except Exception as e:
-        msg = f'Unable to setup SSH for Github and you are trying to connect via SSH: {e}' # noqa
+        msg = f"Unable to setup SSH for Github and you are trying to connect via SSH: {e}"  # noqa
```
AlexIoannides, line 68: Probably don't need the `# noqa` after it has been Blacked?

src/bodywork/git.py
```diff
@@ -44,102 +50,162 @@ def download_project_code_from_repo(
...
+                f" {SSH_PRIVATE_KEY_ENV_VAR} environment variable"
+            ) from e
+
+    try:
+        known_hosts = ssh_dir / "known_hosts"
+        if not known_hosts.exists():
+            known_hosts.touch(0o700, exist_ok=False)
+            known_hosts.write_text(get_ssh_public_key_from_domain(hostname))
+        elif not known_hosts_contains_domain_key(hostname, known_hosts):
+            known_hosts.touch(0o700)
```
AlexIoannides, line 147: Why do you touch it again? You know it exists if you end-up here.

src/bodywork/git.py
```diff
@@ -44,102 +50,162 @@ def download_project_code_from_repo(
...
+
+    :param known_hosts_filepath: path to known_hosts file
+    :param hostname: Host name to check for.
+    :return: bool if the hostname is in the file
+    """
+    return hostname in known_hosts_filepath.read_text()
+
+
+def get_ssh_public_key_from_domain(hostname: str) -> str:
+    """Gets the public SSH key from the host and checks if the fingerprint matches
```
AlexIoannides, line 171: Need to format the docstring (same for the above). I'll see if I can find a linter.

.circleci/config.yml
```diff
@@ -28,36 +28,50 @@ jobs:
...
     executor:
       name: aws-eks/python
       tag: "3.8"
     steps:
       - aws-eks/update-kubeconfig-with-authenticator:
           cluster-name: bodywork-dev
       - checkout
+      - add_ssh_keys:
+          fingerprints:
+            - "3b:3f:26:dc:81:3b:6d:49:88:8e:da:f0:d5:9f:99:d6"
```
AlexIoannides, line 56: Which key does this correspond to?

The one with the fingerprint beginning `e2` is the GitLab one, but I don't recognise the one above.

src/bodywork/git.py
Marios85: That would involve creating a file from the previous command and I wanted to avoid creating a file just for the sake of passing it into the keygen command. I think this is the more robust solution as it removes any file handling.

.circleci/config.yml
Marios85: Yes, this PR isn't ready to review yet so I hadn't cleaned up the config.yml yet.

src/bodywork/git.py
Marios85: 👍

src/bodywork/git.py
Marios85: That's a misunderstanding of the touch command on my part (it doesn't exist on Windows) I thought the mode was also looking at the file access rights but it doesn't.

src/bodywork/git.py
Marios85: Have added a space between the first line and the description otherwise I don't what the docstring requirements are.

.circleci/config.yml
Marios85: It's the Github key (look at SSH keys in CircleCI project settings). 

However I think it is better to change this to the Github Deploy key if it has access to the private repo as well.

tests/unit_and_functional/test_git.py
Marios85: It's Black that's adding the comma's...I can't say I agree with that.

Marios85:
Refactored tests to include a global conftest and also expanded the test teardown to prevent tests affecting each other.

Fixed a Windows OS only bug in Workflow Execution when cleaning up the cloned repo.

.circleci/config.yml
```diff
@@ -25,32 +25,45 @@ jobs:
...
+          command: |
             git config --global user.name "Bodywork Machine Learning Ltd"
             git config --global user.email "info@bodyworkml.com"
       - run:
           name: Installing Python dependencies
           command: pip install -r requirements_dev.txt
       - run: 
           name: Running tests
-          command: tox -e py38-unit_and_functional_tests,py38-static_code_analysis -- -s -vv
+          command: tox -e py38-unit_and_functional_tests,py38-static_code_analysis -- -s -vv --junitxml=~/tox/junit.xml
```
AlexIoannides, line 36: Why do you need to use `--junitxml` to create test results and why are you switching to the output path afterwards?

src/bodywork/constants.py
```diff
@@ -23,19 +23,29 @@
...
-SSH_GITHUB_SECRET_NAME = 'ssh-github-private-key'
+BODYWORK_JOBS_DEPLOYMENTS_SERVICE_ACCOUNT = "bodywork-jobs-and-deployments"
+DEFAULT_LOG_LEVEL = "INFO"
+DEFAULT_LOG_LEVEL_ENV_VAR = "BODYWORK_LOG_LEVEL"
+DEFAULT_PROJECT_DIR = Path("./bodywork_project")
+GIT_SSH_COMMAND = "GIT_SSH_COMMAND"
+PROJECT_CONFIG_FILENAME = "bodywork.yaml"
+SSH_DIR_NAME = ".ssh_bodywork"
+SSH_PRIVATE_KEY_ENV_VAR = "BODYWORK_GIT_SSH_PRIVATE_KEY"
+SSH_SECRET_NAME = "ssh-github-private-key"
```
AlexIoannides, line 41: Should be `SSH_SECRET_NAME = "ssh-git-private-key"`.

Or something like that.

src/bodywork/git.py
```diff
@@ -44,102 +51,166 @@ def download_project_code_from_repo(
...
+        raise RuntimeError("git is not available")
     try:
-        if (get_connection_protocol(url) is ConnectionPrototcol.SSH
-                and get_remote_repo_host(url) is GitRepoHost.GITHUB):
-            setup_ssh_for_github()
-        elif SSH_GITHUB_KEY_ENV_VAR not in os.environ:
-            log.warning('Not configured for use with private GitHub repos')
+        if get_connection_protocol(url) is ConnectionProtocol.SSH:
+            hostname = urlparse(f"ssh://{url}").hostname
+            if hostname is not None and hostname != "":
```
AlexIoannides, line 60: You could abbreviate this line with `if hostname:`.

You were probably not aware, that `None` is cast to `False`, likewise for `""` or any empty list/container.

The section on `truthiness` in [this post](https://blog.jrheard.com/truthiness-and-short-circuit-evaluation-in-python) is worth reading, but the whole article is worth some time at some point.

src/bodywork/git.py
```diff
@@ -44,102 +51,166 @@ def download_project_code_from_repo(
...
-def setup_ssh_for_github() -> None:
+def setup_ssh_for_git_host(hostname: str) -> None:
     """Setup system for SSH interaction with GitHub.
 
     Using the private key assigned to an environment variable, this
     function creates a new SSH configuration in the working directory
     and then tells Git to use it for SSH by exporting the
     GIT_SSH_COMMAND environment variable.
+
+    :param hostname:    Hostname to SSH to.
```
AlexIoannides, line 125: Leading spaces before `Hostname`

src/bodywork/workflow_execution.py
```diff
@@ -273,15 +303,31 @@ def _print_logs_to_stdout(namespace: str, job_or_deployment_name: str) -> None:
...
+        print("-" * 100)
     except Exception:
-        print(f'cannot get logs for {job_or_deployment_name}')
+        print(f"cannot get logs for {job_or_deployment_name}")
+
+
+def remove_readonly(func, path, exc_info):
+    """Error handler for ``shutil.rmtree``.
+
+    If the error is due to an access error (read only file) it attempts to add write
```
AlexIoannides, line 323: line length should be 72 (sorry...).

tests/conftest.py
```diff
@@ -0,0 +1,94 @@
+import shutil
```
AlexIoannides, line 1: I think we need the licensing blurb up-top here.

tests/conftest.py
```diff
@@ -0,0 +1,94 @@
...
+from subprocess import run
+from typing import Iterable
+from bodywork.constants import (
+    SSH_DIR_NAME,
+    GIT_SSH_COMMAND,
+)
+
+
+@fixture(scope="function")
+def test_namespace() -> str:
```
AlexIoannides, line 16: Do we use this anywhere?

tests/integration/test_git_integration.py
```diff
@@ -0,0 +1,92 @@
+"""
```
AlexIoannides, line 1: License text blurb required.

tests/integration/test_git_integration.py
```diff
@@ -0,0 +1,92 @@
...
+Integration Tests for interactions with hosted Git repositories.
```
AlexIoannides, line 2: `tests` not `Tests`.

tests/integration/test_git_integration.py
```diff
@@ -0,0 +1,92 @@
...
+        assert False
+    finally:
+        os.environ.pop(SSH_PRIVATE_KEY_ENV_VAR)
+        if GIT_SSH_COMMAND in os.environ:
+            os.environ.pop(GIT_SSH_COMMAND)
+        if ssh_dir.exists():
+            shutil.rmtree(ssh_dir, onerror=remove_readonly)
+
+
+def remove_readonly(func, path, exc_info):
```
AlexIoannides, line 79: You've got this function in the global `contest.py`. You should be able to import this, if we add `__init__.py` files to the test directories (turning them into proper Python modules and sub-modules, etc.).

AlexIoannides (approved):
Good to go once a few minor changes are made.

.circleci/config.yml
Marios85: For CircleCI to generate the test summary (the tests tab in the build).

src/bodywork/constants.py
Marios85: A rogue github reference!

src/bodywork/git.py
Marios85: Noted and corrected!

tests/conftest.py
Marios85: Yes, it's used heavily in test_k8s_cluster.py.

tests/integration/test_git_integration.py
Marios85: Eagle eye 👁️

tests/integration/test_git_integration.py
Marios85: This function shouldn't have to be in the global conftest.py but it's there because there are integration tests in the unit tests. The real solution is to move those tests out into Integration and/or make them proper unit tests which should be done as part of that ticket or as and when we touch them as part of normal development. 

I suggest leaving this change until then, when it will be removed.

PR #84, opened 2021-05-13
71-expose git commit hash as environment variable in bodywork containers to help with versioning

### LINKED ISSUE #71: Expose Git commit hash as environment variable in Bodywork containers to help with versioning
**Description**
_As a Machine Learning Engineer, I would like access to the Git commit hash of the running pipeline, so that I can use it to tag the artefacts generated in my pipelines._

**Tasks**
- create a new function in the `bodywork.git` module, that uses `git rev-parse --short HEAD` to get the commit hash.
- modify `bodywork.k8s.batch_jobs` and `bodywork.k8s.service_deployments` to take git commit cases as variables as arguments that are then injected as environment variables, in much the same way that GitHub SSH credentials are injected.
- enable `workflow_execution.py` to retrieve and then inject the git commit hash into jobs and deployments.
- add tests and update documentation with a new section along the lines of 'pipeline versioning'.

Marios85:
This PR close #71. The commit hash is retrieved and added to both batch and service stages. 

Some refactoring was also done in `workflow_execution.py` which consisted of breaking down the `run_workflow` method into smaller units.


src/bodywork/workflow_execution.py
```diff
@@ -246,6 +140,166 @@ def run_workflow(
...
+            k8s.delete_job(namespace, job_name)
+            log.info(f"deleted job={job_name} from namespace={namespace}")
+
+
+def _run_service_stages(
+    service_stages: List[ServiceStageConfig],
+    project_name: str,
+    env_vars: k8s.EnvVars,
+    log: Logger,
+    namespace,
```
AlexIoannides, line 204: Missing type hint.

src/bodywork/workflow_execution.py
```diff
@@ -246,6 +140,166 @@ def run_workflow(
...
+    repo_branch,
```
AlexIoannides, line 205: Missing type hint.

AlexIoannides (changes requested):
This is 1st class!

I even learnt something about mocking unit tests in `test_run_workflow_adds_git_commit_to_batch_and_service_env_vars`.

Super-trivial request - there are a couple of missing type definitions. I have no idea how they were allows to slip through...

AlexIoannides:
To prevent the typing issue from happening again, could you add the following to `mypy.ini`:

```ini
[mypy]
...
disallow_untyped_calls=True
disallow_untyped_defs=True
```

I think that should do it.

AlexIoannides:
I've only just noticed that you can click-through on the CodeCov link to get a full report - e.g. [here](https://app.codecov.io/gh/bodywork-ml/bodywork-core/compare/84/diff#diff-c3JjL2JvZHl3b3JrL2dpdC5weQ==). Might be worth a look.

AlexIoannides:
Would you mind adding a quick test for `get_git_commit_hash` raising a `RuntimeError` if the sub-process fails?

Marios85:
I've made all the changes above. I thought you might ask for the exception test for the CodeCov badge. 

The new MyPy rules threw up a lot of errors so I had to fix them in various files if you are wondering why this PR has now touched so many files.

Issue #73, opened 2021-04-20
Send basic usage statistics to a centralised tracking server


AlexIoannides:
**Description**
As a Bodywork Product Manager, I would like to know how Users are using Bodywork, so I can accurately guide Bodywork's roadmap.

**Tasks**
- provide a new configuration parameters - e.g., `logging.usage_stats` - that enables Users to opt-out of usage tracking.
- hard-code the tracking server URL (and possible credentials), as a constant in the `bodywork.constants` module.
- modify `bodywork.workflow_execution.run_workflow` to ping the tracking server when it is called, failing gracefully.
- add a basic test.

PR #89, opened 2021-05-17
73-send basic usage statistics to a centralised tracking server

Marios85:
This PR fulfills the changes required to Bodywork-core for #73. Every time the workflow is executed it will ping the usage trackage server (unless explicitly configured not to) and fail silently if it cannot be reached.

The Constant containing the URL for the tracking server needs to be updated with the tracking server URL once this is available.


src/bodywork/workflow_execution.py
```diff
@@ -83,14 +88,13 @@ def run_workflow(
...
-        log = bodywork_log_factory()
-        log.info(
+        _log.info(
             f"attempting to run workflow for project={repo_url} on "
             f"branch={repo_branch} in kubernetes namespace={namespace}"
         )
         if k8s.namespace_exists(namespace) is False:
             raise ValueError(f"{namespace} is not a valid namespace on your cluster")
-        log.setLevel(config.logging.log_level)
+        _log.setLevel(config.logging.log_level)
```
AlexIoannides, line 97: Good spot.

src/bodywork/workflow_execution.py
```diff
@@ -400,3 +400,19 @@ def _remove_readonly(func: Any, path: Any, exc_info: Any) -> None:
         func(path)
     else:
         raise Exception
+
+
+def _ping_usage_stats_server() -> None:
+    """Pings the usage stats server."""
+    try:
+        param = "-n" if platform.system() == "Windows" else "-c"
+        output = run(
```
AlexIoannides, line 409: I was thinking we'd do something like `requests.get(USAGE_STATS_SERVER_URL)`, so that the tracking server would receive a HTTP request, that is easy to handle with Flask/FastAPI.

Ping uses [Internet Control Message Protocol](https://en.wikipedia.org/wiki/Ping_(networking_utility)) - I have no idea how to work with this on the server-side (scratching at the limits of my knowledge here...).

What I originally had in mind, is repeating what we do for ` bodywork.workflow_execution.image_exists_on_dockerhub`, but without re-raising a `RunTimeError` if we get a `requests.exceptions.ConnectionError`.

What do you think?

src/bodywork/config.py
```diff
@@ -228,6 +229,11 @@ def __init__(self, config_section: Dict[str, str]):
             self.name = config_section["name"]
             self.docker_image = config_section["docker_image"]
             self.DAG = config_section["DAG"]
+            self.usage_stats = (
```
AlexIoannides, line 232: Nice 👍

tests/unit_and_functional/test_workflow_execution.py
```diff
@@ -161,3 +169,56 @@ def test_run_workflow_adds_git_commit_to_batch_and_service_env_vars(
...
+    mock_run: MagicMock,
+    project_repo_location: Path,
+):
+    config_path = Path(f"{project_repo_location}/bodywork.yaml")
+    config = BodyworkConfig(config_path)
+    config.project.usage_stats = False
+
+    run_workflow(config, "foo_bar_foo_993", project_repo_location)
+
+    mock_run.assert_not_called()
```
AlexIoannides, line 224: Need a newline at the end of the file - flake8 should be complaining about that 😬

Is flake8 working post-CICD pipeline refactor?

AlexIoannides (changes requested):
Let me know what you think of ping vs. HTTP GET.

I'm going to Google "Python ping counter" and see if that throws up something nice and simple.

AlexIoannides:
Okay, I can do it with sockets or via installing other software. This is all a bit too low-level for me.

Would you mind swapping to HTTP GET?

src/bodywork/workflow_execution.py
Marios85: I misunderstood the ticket as a literal ping the server.

I don't think a normal http request is appropriate because we don't want to be waiting for the reply, if the server is unreachable it will cause a delay. If we set the timeout request to something short (to avoid waiting for a reply) like 1 second we'll end up throwing exceptions as part of normal process which isn't good.

I think the best option will be to use an async framework for an async request. I know this might seem like overkill for a ping but it's the best implementation if we're not going to use ping itself.

tests/unit_and_functional/test_workflow_execution.py
Marios85: Good spot! I have no idea why Flake8 didn't pick that up as it definitely seems to be running. It has got me on other flake8 errors recently.

Marios85:
I'm cool with that, I think using an async framework is the way to go.

Marios85:
This has now been changed to use an async Http request using the stdlib async package, asyncio.

tests/unit_and_functional/test_workflow_execution.py
```diff
@@ -161,3 +166,46 @@ def test_run_workflow_adds_git_commit_to_batch_and_service_env_vars(
...
+    mock_session: MagicMock,
+    mock_rmtree: MagicMock,
+    project_repo_location: Path,
+):
+    config_path = Path(f"{project_repo_location}/bodywork.yaml")
+    config = BodyworkConfig(config_path)
+
+    run_workflow(config, "foo_bar_foo_993", project_repo_location)
+
+    mock_session().get.assert_called_once()
```
AlexIoannides, line 211: Shouldn't it be checking that it hasn't been called?

src/bodywork/workflow_execution.py
```diff
@@ -400,3 +399,17 @@ def _remove_readonly(func: Any, path: Any, exc_info: Any) -> None:
...
+def _ping_usage_stats_server() -> None:
+    """Pings the usage stats server."""
+    try:
+        session = requests.Session()
+        session.mount(
+            USAGE_STATS_SERVER_URL, requests.adapters.HTTPAdapter(max_retries=0)
+        )
+        response = session.get(USAGE_STATS_SERVER_URL)
+        if not response.ok:
+            _log.warning("Unable to contact usage stats server")
```
AlexIoannides, line 413: Should we not keep schtum here?

tests/unit_and_functional/test_workflow_execution.py
Marios85: It gets called once by the docker image request so we're checking that it's only called once. Unfortunately there isn't a `assert_not_called_with` in pytest.

src/bodywork/cli/cli.py
```diff
@@ -394,7 +392,6 @@ def cronjob(args: Namespace) -> None:
     """Cronjob command handler.
```
AlexIoannides, line 392: Thanks for cleaning these up.

src/bodywork/workflow_execution.py
Marios85: I wasn't sure about this either...could go log.info or just remove it. What do you think?

src/bodywork/cli/cli.py
Marios85: No problem, I naturally clean up as I come across things.

AlexIoannides (approved):
Looks good - just one query around a test and one comment.

src/bodywork/workflow_execution.py
AlexIoannides: I reckon we should use `log.info` for these.

I was about to say, "Log nothing!" But this is probably a bit dishonest and makes it hard to debug, as well.

tests/unit_and_functional/test_workflow_execution.py
AlexIoannides: `not mock_session().get.assert_called_with(USAGE_STATS_SERVER_URL)`?

Marios85: If only it was that easy! The assert is in the mock method call so the test will finish with that `assert_called_with()`.

src/bodywork/workflow_execution.py
Marios85: I agree, log.info sounds like a good compromise. Also don't want to set anyone's logging alarms off with a warning for this.

tests/unit_and_functional/test_workflow_execution.py
AlexIoannides: 🤔

```python
try:
    mock_session().get.assert_called_with(USAGE_STATS_SERVER_URL)
    assert False
except Exception:
    assert True
```

?

Marios85: That looks very hacky. I think it will always return true if a false assertion always creates an exception.

AlexIoannides: I agree...

PR #93, opened 2021-05-23
Delete a deployment if it fails the first time the deployment is attempted

AlexIoannides:
When a deployment is made for the first time, and fails, the workflow will try to roll it back. There is, however, nothing to roll it back to, and so the rollback fails to impact anything. This bug is only observed when resource constraints don't allow the required number of replicas to be scheduled, but would otherwise be successful.

This was fixed, but modifying `bodywork.k8s.service_deployments.rollback_deployment` to delete deployments that have no history - i.e., that are new deployments.


Marios85 (approved):
Perfect! 😎

PR #94, opened 2021-05-23
Add alias for CLI workflow command

AlexIoannides:
If you want to test a workflow locally, you would need to call the `bodywork workflow` command (using positional arguments), whereas to create a remote deployment you would call the `bodywork deployment create` command (with named arguments). 

This has annoyed me for some time now, so I've added a `--test-locally` flag to `bodywork deployment create`, that passes arguments to the `workflow` CLI command hander - i.e. it's an alias, as opposed to a replacement. So,

```
$ bodywork deployment create \
    --namespace=alex \
    --name=initial-deployment \
    --git-repo-url=https://github.com/bodywork-ml/bodywork-pymc3-project \
    --git-repo-branch=main \
    --test-locally
```

Now does the same thing as,

```
$ bodywork workflow \
    --namespace=alex \
    https://github.com/bodywork-ml/bodywork-pymc3-project.git \
    main
```

Which is a much better user experience (in my opinion).


mypy.ini
```diff
@@ -2,7 +2,6 @@
 files = src/bodywork
 warn_return_any = True
 disallow_untyped_calls=True
-disallow_untyped_defs=True
```
Marios85, line 5: Was this showing up errors in test functions for you (despite being set to scan src only)?

Marios85 (approved):
Looks good!

My only suggestion is maybe changing the arg name from test-locally to deploy-local/deploy-locally. Let's not make an assumption as to what the user is doing with that deployment and also to make it absolutely clear that a local deployment is run no different to one on Prod. 
 
 I'm probably reading too much into this so i'm happy with the other option if you don't agree.

mypy.ini
AlexIoannides: Yes and it was doing my f*cking head in.

Marios85: Mine too 😆

AlexIoannides:
The deployment won't be local - it'll still deploy containers to whatever cluster you're point to. It's just that the workflow-controller - `run_workflow` - get run locally. A better description would be, `--local-workflow-controller`. Which is much better.

mypy.ini
AlexIoannides: We that's 'that' decided then!

PR #95, opened 2021-05-24
#92 implement a delete method for deployments

### LINKED ISSUE #92: Implement a delete method for deployments
There is currently no command for,

```
$ bodywork deployment delete \
     --ns NAMESPACE \
     --name DEPLOYMENT_NAME
```

As it was never implemented. It was our intention to rely on the Time To Live (TTL) settings and controller to clean-up finished jobs after they have completed (successfully, or otherwise).

Over time, I've noticed that this function hasn't been working. After some digging, it transpires that this features  of the k8s hasn't been enabled on AWS EKS as it's only in `beta` (where it has been for a few years). So, some cluster may not be able to rely on it, so we should provide a simple method for deleting these deployment jobs...

Tasks
- implement a `delete_workflow_job` method in the `bodywork.cli.workflow_jobs` module; and,
- implement a `bodywork deployment delete` command in the `bodywork.cli.cli` module.

Marios85:
This PR closes #92. A new deployment CLI command `delete_job` has been added that deletes a workflow job (batch job) from a given namespace.


src/bodywork/cli/cli.py
```diff
@@ -92,7 +93,7 @@ def cli() -> None:
     deployment_cmd_parser.add_argument(
         "command",
         type=str,
-        choices=["create", "display", "logs"],
+        choices=["create", "display", "logs", "delete"],
```
AlexIoannides, line 96: I think this needs to be `delete-job` as well.

AlexIoannides (changes requested):
I spotted one small issue, but apart from that it's perfect.

src/bodywork/cli/cli.py
Marios85: Ay, that it does!

PR #100, opened 2021-06-03
#90 provide the option to execute a batch job on workflow failure

### LINKED ISSUE #90: Provide the option to execute a batch job on workflow failure
As a MLOps Engineer, I would like to be able to run custom code (e.g. send notifications) if a workflow has failed to execute, so that I can handle errors appropriately.

**Tasks**
- extend the config file schema with a `project.run_on_failure` parameter, which takes the name of a stage that is only to be run if the workflow raises an exception.
- if an exception was raised during `bodywork.workflow_execution.run_workflow`, then if `project.run_on_failure` has been set, start a job to run the required Python module specific in `project.run_on_failure`, using the exception as an argument to the module.

We should be able to use existing test repos that raise errors, in order to build an integration test for this.

Marios85:
This PR closes #90. 

When configured, the on failure batch job is only run when the error arises from the running of the workflow stages themselves. This is because the prerequisites themselves e.g. valid namespace, Docker image etc, are also required for the running of the 'on failure' batch job. 

A test project repo (bodywork-failing-test-project) was also created for the integration test.


tests/integration/test_k8s_with_cluster.py
```diff
@@ -284,6 +284,32 @@ def test_workflow_will_cleanup_jobs_and_rollback_new_deployments_that_yield_erro
...
+                "https://github.com/bodywork-ml/bodywork-failing-test-project",  # noqa
+                "master",
+                f"--bodywork-docker-image={docker_image}",
+            ],
+            encoding="utf-8",
+            capture_output=True,
+        )
+
+        expected_output_0 = "ERROR - workflow_execution.run_workflow - failed to execute workflow for master branch"  # noqa
+        expected_output_1 = "successfully ran stage=on_fail_stage"
```
AlexIoannides, line 305: You could print to stdout from the on-failure-stage and look for that message, as well, as an `expected_output_3`?

tests/unit_and_functional/test_workflow_execution.py
```diff
@@ -168,6 +170,116 @@ def test_run_workflow_adds_git_commit_to_batch_and_service_env_vars(
...
+@patch("bodywork.workflow_execution.requests.Session")
+@patch("bodywork.workflow_execution.k8s")
+def test_failure_stage_does_not_run_for_docker_image_exception(
+    mock_k8s: MagicMock, mock_session: MagicMock, project_repo_location: Path
+):
+    config_path = Path(f"{project_repo_location}/bodywork.yaml")
+    config = BodyworkConfig(config_path)
+    response = requests.Response()
+    response.status_code = 401
+    mock_session().get.return_value = response
```
AlexIoannides, line 230: `mock_session().get.return_value = requests.Response()` would be cleaner?

tests/unit_and_functional/test_workflow_execution.py
```diff
@@ -168,6 +170,116 @@ def test_run_workflow_adds_git_commit_to_batch_and_service_env_vars(
...
+    mock_git_download: MagicMock,
+    mock_requests: MagicMock,
+    mock_rmtree: MagicMock,
+    project_repo_location: Path,
+):
+    config_path = Path(f"{project_repo_location}/bodywork.yaml")
+    config = BodyworkConfig(config_path)
+    config.project.run_on_failure = "on_fail_stage"
+
+    error_message = "Failure Stage Error"
```
AlexIoannides, line 272: I think that the error message might be clearer if it were `The run-on-failure stage experienced an error`. I'm concerned that `failure stage` might not register with some people casually looking through stuff.

AlexIoannides (approved):
This is **excellent** - 1st class quality work...

I just have a few very (very) minor comments around documenting the functions that raise the new exceptions and the error message for a failed on-failure stage.

src/bodywork/workflow_execution.py
```diff
@@ -91,7 +97,9 @@ def run_workflow(
             f"branch={repo_branch} in kubernetes namespace={namespace}"
         )
         if k8s.namespace_exists(namespace) is False:
-            raise ValueError(f"{namespace} is not a valid namespace on your cluster")
+            raise BodyworkNamespaceError(
```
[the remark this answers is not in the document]
Marios85: I'm don't know what you mean by the exception type but I assume it's the fact that namespace_exists could return an API Error.

I'll wrap that call in a try/except block to keep the k8s.api functions consistent by throw k8s errors.

tests/integration/test_k8s_with_cluster.py
Marios85: Yeah I can easily add that in, it already is printing something to stdout.

tests/unit_and_functional/test_workflow_execution.py
Marios85: I agree!

tests/unit_and_functional/test_workflow_execution.py
Marios85: Yeah I can see how that could be confusing.

src/bodywork/workflow_execution.py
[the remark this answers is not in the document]
Marios85: You mean documentation? This function will still return a BodyworkworkflowExecution error but I see that the private methods for the docker ones are missing this.

Marios85:
Thanks for the review! I'll make the changes and then put this into a new PR to go into a V2.1.1 branch.

PR #104, opened 2021-06-10
V2.1.2

Marios85:
Ping usage stats server on install so that we get accurate figures for how many people install Bodywork.


AlexIoannides (approved):
LGTM!

PR #105, opened 2021-06-12
V2.1.3 - SSH Key validation, run workflow refactor & Fixes

Marios85:
This PR contains the following:

- Add a new line to the end of the SSH Key environment variable if it is missing so that the key is valid. This is to prevent likely user error from copying and pasting the key into their environment variable. 
- Stop cloned repo cleanup throwing an error if the clone failed and there is no repo to clean up.
- Refactor run workflow so that the config is retrieved within the workflow and can be overridden. This also fixes a potential issue when working with remote repos when the git commit has is taken after the repo is cleaned up or locally.


src/bodywork/exceptions.py
```diff
@@ -38,19 +38,23 @@ def __init__(self, failed_jobs: Iterable[V1Job]):
         super().__init__(msg)
 
 
-class BodyworkConfigParsingError(Exception):
+class BodyworkConfigError(Exception):
```
AlexIoannides, line 41: I'm not sure why this exists - it's acting as a wrapper for `Exception`. Superfluous, no?

src/bodywork/workflow_execution.py
```diff
@@ -42,53 +42,34 @@
...
     :param namespace: Kubernetes namespace to execute the workflow in.
     :param repo_url: Git repository URL.
     :param repo_branch: The Git branch to download, defaults to 'master'.
     :param docker_image_override: Docker image to use for executing all
         stages, that will override the one specified in the
         project config file. Provided purely for testing purposes and
         defaults to None.
+    :param config: Configuration data for the Bodywork deployment.
+    :param cloned_repo_dir: The name of the directory into which the
+    repository will be cloned, defaults to DEFAULT_PROJECT_DIR.
```
AlexIoannides, line 72: I think this line needs an indent?

src/bodywork/workflow_execution.py
```diff
@@ -167,24 +152,32 @@ def run_workflow(
...
+                    BodyworkConfigError,
+                ]
+            ):
                 _run_failure_stage(
                     config, e, namespace, repo_url, repo_branch, docker_image
                 )
         except Exception as ex:
             failure_msg = (
-                f"Error executing failure stage: {config.project.run_on_failure}"
+                f"Error executing failure stage: {config.project.run_on_failure}"   # type: ignore # noqa
```
AlexIoannides, line 171: Why do you need to ignore the type - this is usually a bad smell?

src/bodywork/workflow_execution.py
```diff
@@ -167,24 +152,32 @@ def run_workflow(
         )
         _log.error(msg)
         try:
-            if config.project.run_on_failure and type(e) not in [
-                BodyworkNamespaceError,
-                BodyworkDockerImageError,
-                BodyworkGitError,
-            ]:
+            if (
+                config is not None
```
AlexIoannides, line 156: Why do you need this? 

Are you getting shitty mypy errors?

src/bodywork/workflow_execution.py
```diff
@@ -167,24 +152,32 @@ def run_workflow(
...
                 f" after failed workflow : {ex}"
             )
             _log.error(failure_msg)
             msg = f"{msg}\n{failure_msg}"
         raise BodyworkWorkflowExecutionError(msg) from e
     finally:
-        if config.project.usage_stats:
+        if cloned_repo_dir.exists():
+            rmtree(cloned_repo_dir, onerror=_remove_readonly)
+        if config is not None and config.project.usage_stats:
```
AlexIoannides, line 180: Related to the above - config cannot be none at this stage.

src/bodywork/workflow_execution.py
```diff
@@ -429,8 +422,8 @@ def parse_dockerhub_image_string(image_string: str) -> Tuple[str, str]:
 def _print_logs_to_stdout(namespace: str, job_or_deployment_name: str) -> None:
     """Replay pod logs from a job or deployment to stdout.
 
-    :param namespace: The namspace the
-    :param job_or_deployment_name: THe name of the pod or deployment.
+    :param namespace: The namespace the job/deployment is in.
+    :param job_or_deployment_name: The name of the pod or deployment.
```
AlexIoannides, line 426: I love your attention to detail.

tests/integration/test_git_integration.py
```diff
@@ -101,7 +101,7 @@ def test_setup_ssh_for_git_host_create_known_host_and_env_var():
 
         private_key = ssh_dir / "id_rsa"
         assert private_key.exists() is True
-        assert private_key.read_text() == "MY_PRIVATE_KEY"
+        assert private_key.read_text() == "MY_PRIVATE_KEY\n"
```
AlexIoannides, line 104: Is it worth adding in a version of this test where,

```python
os.environ[SSH_PRIVATE_KEY_ENV_VAR] = "MY_PRIVATE_KEY\n"
```

Would this be overkill?

src/bodywork/git.py
```diff
@@ -240,3 +242,7 @@ def get_git_commit_hash(project_path: Optional[str] = None) -> str:
         raise BodyworkGitError(
             f"Unable to retrieve git commit hash: {e.stdout} {e.stderr}"
         ) from e
+    except OSError as e:
```
AlexIoannides, line 245: Can we have a test for this path?

src/bodywork/workflow_execution.py
```diff
@@ -42,53 +42,34 @@
...
-    return config
-
-
 def run_workflow(
-    config: BodyworkConfig,
     namespace: str,
     repo_url: str,
     repo_branch: str = "master",
     docker_image_override: Optional[str] = None,
+    config: Optional[BodyworkConfig] = None,
```
AlexIoannides, line 58: if you rename this `config_override` then,

```python
        config = (
            config_override
            if config_override is not None
            else BodyworkConfig(cloned_repo_dir / PROJECT_CONFIG_FILENAME, True)
        )
```

Should solve all you problems (have checked it at my end).

src/bodywork/workflow_execution.py
AlexIoannides: See comment above.

src/bodywork/workflow_execution.py
AlexIoannides: See comment above.

src/bodywork/workflow_execution.py
AlexIoannides: I you don't like the name `config_override` then you could use `config` in the argument and `config_` within the body of the function?

AlexIoannides: I you don't like the name `config_override` then you could use `config` in the argument and `config_` within the body of the function?

src/bodywork/workflow_execution.py
```diff
@@ -462,7 +455,7 @@ def _remove_readonly(func: Any, path: Any, exc_info: Any) -> None:
         os.chmod(path, stat.S_IWRITE)
         func(path)
     else:
-        raise Exception
+        _log.warning(f"Could not remove file/directory: {path}")
```
AlexIoannides, line 458: It's an error, isn't it?

Getting tired now, so I may not be thinking straight.

AlexIoannides (changes requested):
A few comments here and there, but the overall logic is sound.

src/bodywork/exceptions.py
Marios85: It's a base exception class for Config errors so that when we check the exception type in run_workflow we don't need to check each one so its a lot cleaner.

src/bodywork/workflow_execution.py
Marios85: Good Spot!

src/bodywork/workflow_execution.py
Marios85: 🙂

src/bodywork/workflow_execution.py
Marios85: This and your following comments are all related to MyPy errors. The config can't logically be null but the MyPy check can't calculate it.

tests/integration/test_git_integration.py
Marios85: Testing an individual 'If' statement is overkill. Our other tests will fail if it cannot handle the '/n' at the end as they all read from the file programatically and therefore have the line break at the end.

src/bodywork/git.py
Marios85: I'm guessing this is for the sake of the CodeCov report.

src/bodywork/workflow_execution.py
Marios85: If this fixes the false positives from MyPy then i'm for it. That's poor form for Python that the type checker doesn't work properly but I suppose that's what you get with a language that isn't strongly typed.

src/bodywork/workflow_execution.py
Marios85: This is just used for cleaning up the cloned repo directory at the end of the workflow therefore if it can't do it for whatever reason it shouldn't result in an otherwise successful workflow, reporting as failed (non zero exit code).

src/bodywork/git.py
Marios85: Done

src/bodywork/exceptions.py
AlexIoannides: Understood - good idea.

src/bodywork/git.py
AlexIoannides: Nice one thanks.

src/bodywork/workflow_execution.py
AlexIoannides: MyPy is still relatively new and is constantly being updated. Guido von Rossum looks after the project personally.

src/bodywork/workflow_execution.py
AlexIoannides: Agreed.

tests/integration/test_git_integration.py
AlexIoannides: You're right.

AlexIoannides (approved):

PR #110, opened 2021-06-23
Fix MyPy complaints

AlexIoannides:
Some people like to solve crossword puzzles; I like to fix MyPy bugs.

Let me know what you think.


Marios85 (approved):
It does the job but its wrong that Mypy should make us repeat code to workaround its shortcomings. Its reduced our codecov as well which is annoying (although im cool with that).

AlexIoannides:
With the benefit of hindsight, we could have done, 

```python
config: Optional[BodyworkConfig] = None
...
```

As opposed to,

```python
config = None
...
```

So that MyPy would know our true intentions for `config`. That said, the flow we've ended-up with is more intuitive and easier to follow (in my opinion). I really don't think MyPy is at fault here.

Marios85:
That would still result in the same Mypy errors because Mypy cannot deal with the short circuiting in the if statement.

I disagree the flow is less intuitive, a single if statement in a finally block shows clearly that the usage stats is run at the end of a workflow regardless of anything. Having two if statements means there are two logical flows to look at and it also means it does not adhere to the DRY principle as well either.

So I guess we'll have to agree to disagree on this one.

AlexIoannides:
It's an interesting one.

You an end-up without a config, if fetching the repo fails. In this instance, you don't know whether or not you have permission from the user to send usage stats. So the currently flow, only does this when we have this permission, and there are two paths in this branch (successful or failed workflow), which we handle. If we handle this in a finally block, we have to manage the possibility that we ended up there without a config. I just prefer the former over the latter. No doubt, there's an even clearer version that we haven't thought of yet!

AlexIoannides:
> _"That would still result in the same Mypy errors because Mypy cannot deal with the short circuiting in the if statement."_

You're right (just tested this). I think we're hitting the limits of static type checkers. What's it like with TypeScript?

AlexIoannides:
Take a look at [this thread](https://github.com/python/mypy/issues/4805). I think this is a common grievance, but it's worth noting that there appears to be an option to 'relax' the checking for optionals like this.

Marios85:
I didnt use TypeScript with React for my app so I don't know what its like. 

A bug from 2018 that is still not addressed 😐  If you can relax it for an individual file then that would be the best option.

AlexIoannides:
That's not a bug - that's a limitation.

Marios85:
Its giving false positives therefore its a bug in my opinion.

PR #113, opened 2021-06-27
111 create a configure cluster command

### LINKED ISSUE #111: Create a `configure cluster` command
_As a Machine Learning Engineer, I would like all k8s cluster setup to be done for me, so that I don't need to understand k8s before deploy my pipelines._

**Tasks**
- Create a `bodywork.cli.configure_cluster` module.
- Implement a `setup_cluster_for_bodywork` function that will perform all the necessary steps to configure k8s for use with bodywork - e.g.,
  - create a `bodywork-deployment-jobs` namespace, into which **all** deployment (workflow) jobs will be run.
  - create service accounts and roles for `bodywork-deployment-jobs`, so that workflow-jobs will be able to create resources. This could be done with [bodywork.k8s.auth.setup_workflow_service_account](https://github.com/bodywork-ml/bodywork-core/blob/2f2a0b7e44cb648d164445b4717d4e6ef4bd8385/src/bodywork/k8s/auth.py#L115), but it may have to be modified with the option to grant permission for namespaces to be created, which will be required by workflow-jobs in `bodywork-deployment-jobs` (as we want to automate namespace creation).

Marios85:
This PR closes #111. Tweaked the existing code for setting up a namespace so the job-deployment accounts have permission to create namespaces. This in turn is called by the new command to configure the cluster.


src/bodywork/k8s/auth.py
```diff
@@ -229,7 +229,11 @@ def setup_job_and_deployment_service_accounts(namespace: str) -> None:
                 api_groups=[""],
                 resources=["secrets", "configmaps"],
                 verbs=["get", "list"],
-            )
+            ),
+            k8s.V1PolicyRule(
```
AlexIoannides, line 233: I think this needs to go in `setup_workflow_service_account`.

This creates the service accounts that will be used by batch and service stages, so we should not authorise them to be able to interact with anything at the cluster level.

In writing this, I realise how unclear the naming is - sorry.

tests/unit_and_functional/test_cli.py
```diff
@@ -371,6 +371,12 @@ def test_debug_subcommand_sleeps():
     assert process.returncode == 0
 
 
+def test_configure_cluster_subcommand_exists():
```
AlexIoannides, line 374: I wonder how best to test this functionality, given that we can only configure our test cluster once 🤔

Maybe we tear-down the configuration after tests have run and require `configure-cluster` for every integration test?

AlexIoannides (changes requested):
A couple of points - happy to jump on a call to discuss.

src/bodywork/cli/cli.py
```diff
@@ -605,3 +610,13 @@ def validate_config(args: Namespace) -> None:
         missing_or_invalid_param_list = "\n* ".join(e.missing_params)
         print(f"* {missing_or_invalid_param_list}")
         sys.exit(1)
+
+
+def configure_cluster(args: Namespace):
```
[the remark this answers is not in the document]
Marios85: Completely missed that!

src/bodywork/constants.py
```diff
@@ -31,6 +31,7 @@
 BODYWORK_WORKFLOW_SERVICE_ACCOUNT = "bodywork-workflow-controller"
 BODYWORK_WORKFLOW_JOB_TIME_TO_LIVE = 15 * 60
 BODYWORK_JOBS_DEPLOYMENTS_SERVICE_ACCOUNT = "bodywork-jobs-and-deployments"
+BODYWORK_JOBS_DEPLOYMENTS_NAMESPACE = "bodywork-jobs-and-deployments"
```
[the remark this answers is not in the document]
Marios85: Yeah that makes more sense to me.

src/bodywork/k8s/auth.py
Marios85: Yeah I got confused here. I am very glad you spotted it. Ive made the change.

tests/unit_and_functional/test_cli.py
Marios85: I think the best approach will be to set it up once for all the tests in a test fixture setup. It doesnt make sense to spin this up every time as its stateless and none of our test should modify any aspect of it. Also it will make all tests run longer and eat into our CICD credit. 

Right now the setup namespace tests cover this and with the next ticket where we remove the namespace from other functions it will be tested even more.

AlexIoannides: Agreed.

src/bodywork/constants.py
```diff
@@ -30,8 +30,8 @@
 BODYWORK_WORKFLOW_CLUSTER_ROLE = "bodywork-workflow-controller"
 BODYWORK_WORKFLOW_SERVICE_ACCOUNT = "bodywork-workflow-controller"
 BODYWORK_WORKFLOW_JOB_TIME_TO_LIVE = 15 * 60
-BODYWORK_JOBS_DEPLOYMENTS_SERVICE_ACCOUNT = "bodywork-jobs-and-deployments"
-BODYWORK_JOBS_DEPLOYMENTS_NAMESPACE = "bodywork-jobs-and-deployments"
+BODYWORK_DEPLOYMENT_JOBS_SERVICE_ACCOUNT = "bodywork-deployment-jobs"
```
AlexIoannides, line 33: We're still going to need a service account for the jobs and deployments (stages) that are deployed into a namespace.

The I think needs to  be renamed is `BODYWORK_WORKFLOW_SERVICE_ACCOUNT`.

Sorry, this is my crappy naming that's causing the confusion.

AlexIoannides (changes requested):
I think the logic is good-to-go - we just need to get the naming sorted so that it makes sense.

We have two sets of entity that required service-accounts:
- pods running the workflow-controller (i.e. in a 'deployment job'); and,
- pods that form the stages of a workflow (be them jobs or k8s-deployment resources).

Prior to this PR, we have `BODYWORK_WORKFLOW_SERVICE_ACCOUNT` for the first entity and `BODYWORK_JOBS_DEPLOYMENTS_SERVICE_ACCOUNT` for the second.

We need to pick names that done't confuse in the future.

src/bodywork/constants.py
Marios85: Im confused now, BODYWORK_WORKFLOW_SERVICE_ACCOUNT already exists and it is used for the workflow controller.

AlexIoannides: Correct.

When you run `bodywork deployment create` it creates a k8s job, that starts a pod, that runs the `bodywork workflow` command within a container.

So, this pod needs a service account that can access the k8s namespace API. Currently, this is `BODYWORK_WORKFLOW_SERVICE_ACCOUNT`. This is the one that needs to be modified to be able to create and delete namespaces, etc.

The stages that are deployed by the workflow controller, also need to have a custom service account so that they don't take the `default` service account that exists in every namespace (which is too permissive). This used to be called `BODYWORK_JOBS_DEPLOYMENTS_SERVICE_ACCOUNT`.

I have probably confused things, by referring to k8s jobs running the workflow-controller as 'deployment jobs'.

Marios85: I still dont know what constant you want renamed because BODYWORK_WORKFLOW_SERVICE_ACCOUNT already exists and all i have done is rename BODYWORK_JOBS_DEPLOYMENTS_SERVICE_ACCOUNT to BODYWORK_DEPLOYMENT_JOBS_SERVICE_ACCOUNT. Obviously I cannot rename this the same as another constant and I dont see why you would want to either.

AlexIoannides: Sorry this is all my fault. How about,

`BODYWORK_WORKFLOW_SERVICE_ACCOUNT` is left alone.

And,

`BODYWORK_JOBS_DEPLOYMENTS_SERVICE_ACCOUNT`  --> `BODYWORK_STAGES_SERVICE_ACCOUNT`

Less ambiguity this way?

Marios85:
Renaming has been done and added a test fixture that configures the cluster for the tests that require it.

AlexIoannides:
`setup_workflow_service_account(...)` needs a little modification here,

```python
k8s.V1PolicyRule(
    api_groups=[""], resources=["namespaces"], verbs=["get", "list"]
),

So that the workflow-controller can also `create` and `delete`.
```

tests/integration/conftest.py
```diff
@@ -135,3 +139,14 @@ def ingress_load_balancer_url() -> str:
         raise RuntimeError(msg)
     except Exception as e:
         raise RuntimeError() from e
+
+
+@fixture(scope="session")
+def setup_cluster(request):
```
AlexIoannides, line 145: Do you mind adding a type - hint? 

Use the search on the pytest docs to dig it up. This is new functionality for me and I have no ideas what this thing is 🙂

AlexIoannides (changes requested):
Just a couple of minor point and then good to go - doesn't need another review.

tests/integration/conftest.py
Marios85: Sure, I admit it's not obvious at all where it comes from.

Marios85:
I have made the changes.

I didn't add the `delete` command as I thought that the user would be issuing the command to delete a namespace but then I suppose the workflow controller might want to tear down the namespace on fail etc.

AlexIoannides:
Yes, that's right - if a deployment doesn't leave any services hanging around (i.e. just batch jobs), we should probably delete the namespace, too.

PR #114, opened 2021-07-04
Hotfix - add PYTHONPATH to `run_stage` environment and handle missing config file exceptions

AlexIoannides:
This hotfix targets two separate issues:
1. While putting together an [example pipeline project](https://github.com/bodywork-ml/ml-pipeline-engineering) for the next blog post, I run into several issues when using a `pipeline/` folder for the pipeline code and then trying to import modules as if `pipeline` was a Python package module. These can be solved by adding the project's root directory to `PYTHONPATH`. While this isn't strictly necessary, it leads to a **much** better developer experience (and probably reduces the risk of confusion).
- By accident, I tried to run a workflow without a `bodywork.yaml` file and I got a cascade of exceptions in the face. This is because `BodyworkConfig` throws a `FileExistsError` exception, which we don't handle in `run_workflow`. I thought about creating a custom exception that inherits from `BodyworkConfigError`, but that seemed overkill (need your opinion here).


Marios85 (changes requested):
For consistency that should be covered as a BodyworkConfig error as all other config errors are raised as such. I do not think there is a need to create a custom exception and just raising it as the Base Exception is sufficient.

AlexIoannides:
I decided to write a custom exception in the end...

Marios85 (approved):
All good!

AlexIoannides:
Thank you!

PR #117, opened 2021-07-15
112 automate the management of k8s namespaces

### LINKED ISSUE #112: Automate the management of k8s namespaces
_As a Machine Learning Engineer, I would like the creation and specification of k8s namespaces to be handled automatically by Bodywork, so that I do not have to know about k8s namespaces or what to do with them (I find this all very intimidating)._

**Tasks**
- The user should be provided with the **option** to specify a namespace in `bodywork.yaml` - e.g. in a `project.namespace` field - but if this isn't provided, it should default to `project.name`.
- All remote workflow-jobs should be deployed to the `bodywork-deployment-jobs` namespace, which will be created automatically when the cluster is configured for Bodywork. This means that the `--namespace` flag can be dropped for all `bodywork deployment` commands.
- We should remove the need to specify a deployment name with the `--name` flag and instead use one _based_ on the git-repo URL and the current timestamp, so that the resulting command looks like `bodywork deployment create MY_REPO MY_BRANCH`.
- Refactor `bodywork.workflow_execution.run_workflow`, so that it doesn't take `namespace` as an argument and instead creates the namespace, if it needs to, based on the config parameters. If the workflow does not deploy any services, then the namespace should be deleted when the workflow has successfully completed.
- Refactor the code in `bodywork.k8s.workflow_jobs` and `bodywork.cli.cli.workflow`, to reflect the fact that `namespace` no longer needs to be thrown around as an argument.

Marios85:
This PR closes #112 .

Implementation differs from the original issue requirements in the following way:

- Job name is post-fixed with a small random number instead of a timestamp because there is a 52 character limit on the job name.
- Cronjobs will have a static name as defined by the user as this best fits how people like to use cronjobs. 
- Namespace left as a parameter in lower level package functions as I see no gain in removing them and might come useful in later iterations.


src/bodywork/cli/workflow_jobs.py
```diff
@@ -72,8 +73,7 @@ def delete_workflow_job_in_namespace(namespace: str, job_name: str) -> None:
     print(f"workflow job={job_name} deleted from namespace={namespace}")
 
 
-def create_workflow_cronjob_in_namespace(
-    namespace: str,
+def create_workflow_cronjob(
     schedule: str,
     project_name: str,
```
AlexIoannides, line 78: We pass `--name` into this field, which confused the hell out of me. I take full responsibility, but would you mind tidying-up the naming and docstring?

src/bodywork/k8s/auth.py
```diff
@@ -209,8 +210,8 @@ def setup_workflow_service_account(namespace: str) -> None:
         )
 
 
-def setup_job_and_deployment_service_accounts(namespace: str) -> None:
-    """Setup a jobs-and-deployments service-account with required roles.
+def setup_job_and_deployment_service_account(namespace: str) -> None:
```
AlexIoannides, line 213: Maybe we should refactor this to `setup_stages_service_account`. My own naming choices are confusing me!

src/bodywork/k8s/workflow_jobs.py
```diff
@@ -97,6 +103,18 @@ def configure_workflow_job(
...
 
+def _create_job_name(project_repo_url: str, project_repo_branch: str) -> str:
+    """Create unique job name.
+
+    :param project_repo_url: The URL for the Bodywork project Git
+        repository.
+    :param project_repo_branch: The Bodywork project Git repository
+        branch being used.
+    """
+    repo_name = os.path.splitext(os.path.basename(project_repo_url))[0]
```
AlexIoannides, line 114: Might be easier with a regex (re module)?

src/bodywork/k8s/workflow_jobs.py
```diff
@@ -141,12 +156,12 @@ def configure_workflow_cronjob(
     :return: A configured k8s cronjob object.
     """
     job = configure_workflow_job(
-        namespace=namespace,
-        project_name=project_name,
+        namespace=BODYWORK_DEPLOYMENT_JOBS_NAMESPACE,
```
AlexIoannides, line 159: Should we not have `namespace` as an arg to `configure_workflow_cronjob` as well, for the sake of consistency (that being that namespace is an arg at the low-level k8s API and set in stone at the CLI level)?

src/bodywork/workflow_execution.py
```diff
@@ -51,23 +52,21 @@
 
 
 def run_workflow(
-    namespace: str,
```
AlexIoannides, line 54: Should we not keep this here and set it to `BODYWORK_DEPLOYMENT_JOBS_NAMESPACE` at the CLI level.

src/bodywork/workflow_execution.py
```diff
@@ -109,6 +97,7 @@ def run_workflow(
         env_vars = k8s.create_k8s_environment_variables(
             [(GIT_COMMIT_HASH_K8S_ENV_VAR, get_git_commit_hash(cloned_repo_dir))]
         )
+        workflow_has_services = False
```
AlexIoannides, line 100: We could answer this directly via `config`?

I have an aversion to mutating variables like this, because it makes the code harder to follow.

src/bodywork/workflow_execution.py
```diff
@@ -183,6 +177,31 @@ def run_workflow(
             rmtree(cloned_repo_dir, onerror=_remove_readonly)
 
 
+def _setup_namespace(config) -> str:
+    """Creates namespace to run workflow in.
+
+    :param config: Bodywork config.
+    :return: Name of namespace.
+    """
+    namespace = ""
```
AlexIoannides, line 186: `namespace = config.project.namespace is config.project.namespace else config.project.name`

AlexIoannides: Is MyPy being a PITA again?

tests/unit_and_functional/test_cli.py
```diff
@@ -206,11 +206,10 @@ def test_deployment_subcommand_exists():
 def test_deployment_test_locally_option_calls_run_workflow_handler(
```
AlexIoannides, line 206: Rename to `test_deployment_run_locally_option_calls_run_workflow_handler`, because this is another p*ss poor name from me...

tests/unit_and_functional/test_cli.py
```diff
@@ -219,7 +218,7 @@ def test_deployment_test_locally_option_calls_run_workflow_handler(
     )
     deployment(args)
     expected_pass_through_args = Namespace(
-        namespace="foo1",
+        namespace="bodywork-deployment-jobs",
```
AlexIoannides, line 221: We shouldn't need this anymore, because we shouldn't be passing it through.

src/bodywork/cli/cli.py
```diff
@@ -371,25 +350,23 @@ def deployment(args: Namespace) -> None:
...
         print("please specify Git repo URL for the deployment you want to create")
         sys.exit(1)
+    if command != "create" and name == "":
+        print("please specify --name for the job")
+        sys.exit(1)
     if command == "create":
         if run_workflow_controller_locally:
             pass_through_args = Namespace(
-                namespace=namespace,
+                namespace=BODYWORK_DEPLOYMENT_JOBS_NAMESPACE,
```
AlexIoannides, line 369: I don't think we need to pass this through anymore as it's not used in `workflow`?

tests/unit_and_functional/test_cli.py
```diff
@@ -231,34 +230,24 @@ def test_deployment_test_locally_option_calls_run_workflow_handler(
 
 def test_cli_deployment_handler_error_handling():
     process_one = run(
-        ["bodywork", "deployment", "create", "--namespace=bodywork-dev"],
+        ["bodywork", "deployment", "logs"],
         encoding="utf-8",
         capture_output=True,
     )
-    assert "please specify --name for the deployment" in process_one.stdout
+    assert "please specify --name for the job" in process_one.stdout
```
AlexIoannides, line 237: It should still be `deployment` or `deployment-job`? Otherwise it might get confusing.

tests/unit_and_functional/test_workflow_execution.py
```diff
@@ -347,7 +354,55 @@ def test_usage_stats_opt_out_does_not_ping_usage_stats_server(
...
+@patch("bodywork.workflow_execution.k8s")
+def test_namespace_is_not_deleted_if_there_are_service_stages(
+    mock_k8s: MagicMock,
+    mock_git_hash: MagicMock,
+    mock_git_download: MagicMock,
+    mock_requests: MagicMock,
+    mock_rmtree: MagicMock,
+    project_repo_location: Path,
+):
+    config_path = Path(f"{project_repo_location}/bodywork.yaml")
```
AlexIoannides, line 376: `"{project_repo_location}/{PROJECT_CONFIG_FILENAME}"`

tests/unit_and_functional/test_workflow_execution.py
```diff
@@ -18,7 +18,7 @@
 Test Bodywork workflow execution.
 """
 from pathlib import Path
-from unittest.mock import MagicMock, patch, ANY
+from unittest.mock import MagicMock, patch, ANY, PropertyMock
```
AlexIoannides, line 21: `PropertyMock` isn't used anywhere.

I was about to ask you to teach me about it...

src/bodywork/cli/workflow_jobs.py
Marios85: This was bugging me too but I didn't want to disrupt the order of things as it's used in a few places.

Marios85: I'll change it in the other places too.

tests/integration/test_k8s_with_cluster.py
```diff
@@ -46,28 +45,25 @@
 
 @mark.usefixtures("setup_cluster")
```
AlexIoannides, line 46: Why have you gone down the `@mark.usefixtures("setup_cluster")` route for fixture, rather than how we've been doing them up to this time?

This is for my own understanding.

tests/integration/test_k8s_with_cluster.py
```diff
@@ -46,28 +45,25 @@
...
 def test_workflow_and_service_management_end_to_end_from_cli(
-    random_test_namespace: str, docker_image: str, ingress_load_balancer_url: str
+ docker_image: str, ingress_load_balancer_url: str
 ):
     try:
-        sleep(5)
-
         process_zero = run(
-            ["bodywork", "setup-namespace", random_test_namespace],
+            ["bodywork", "setup-namespace", "bodywork-test-project"],
```
AlexIoannides, line 52: This means we can't both be running tests at the same time. Is this interim or something we need to live with for the time being?

src/bodywork/k8s/auth.py
Marios85: Agreed!

tests/integration/test_k8s_with_cluster.py
AlexIoannides: I'm referring to making the namespace deterministic.

src/bodywork/k8s/workflow_jobs.py
Marios85: I find this a lot more readable than the regex solution.

src/bodywork/k8s/workflow_jobs.py
Marios85: Yes I agree completely, I don't know why I did this one differently.

tests/integration/test_k8s_with_cluster.py
```diff
@@ -522,42 +497,13 @@ def test_cli_secret_handler_crud(test_namespace: str):
...
-            "--git-repo-url=https://github.com/bodywork-ml/bodywork-test-project",
-        ],
-        encoding="utf-8",
-        capture_output=True,
-    )
-    assert f"{random_test_namespace} is not setup for use" in process_one.stdout
-    assert process_one.returncode == 1
-
-
 def test_deployment_of_remote_workflows(random_test_namespace: str):
```
AlexIoannides, line 500: Doesn't need `random_test_namespace` now?

tests/integration/test_k8s_with_cluster.py
AlexIoannides: I see what you're doing now - very clever!

Pre-setup the namespace the project is deployed to, with the secrets loaded into it, until we sort secrets out.

NICE ONE!

AlexIoannides (changes requested):
This is really good - just a few bits and pieces here and there.

src/bodywork/cli/workflow_jobs.py
AlexIoannides: Brilliant - thank you!

src/bodywork/workflow_execution.py
Marios85: I know what you mean but I think the config solution will mean creating another nested for loop just for this as you need to check both the DAG and the stage list.

src/bodywork/workflow_execution.py
Marios85: Yup, pure MyPy PITA

src/bodywork/k8s/workflow_jobs.py
AlexIoannides: I've just tested this...

... and I agree 🙂

tests/unit_and_functional/test_cli.py
Marios85: It's just a typo, don't be so hard on yourself.

tests/unit_and_functional/test_cli.py
Marios85: Yup, that evaded the cleanup as did the one below

src/bodywork/workflow_execution.py
AlexIoannides: I think the suggestion above should solve it, maybe `namespace: str = ...` is required.

tests/unit_and_functional/test_cli.py
Marios85: We're mixing terminology too much from between deployment, workflow, job etc. I think deployment job sounds the clearest to me.

tests/unit_and_functional/test_cli.py
AlexIoannides: It's never a good thing when you don't understand what it is you did a while back...

tests/unit_and_functional/test_cli.py
Marios85: I take that back, let's keep it deployment. We keep it deployment to the user and it's just us who will refer to it as a job.

tests/unit_and_functional/test_workflow_execution.py
Marios85: Good spot! I did use it and then didn't need it anymore.

tests/integration/test_k8s_with_cluster.py
Marios85: Because we don't need access to the fixture object as there is none therefore we want to avoid the requirement of passing a parameter to the test method.

src/bodywork/workflow_execution.py
AlexIoannides:
```python
def config_deploys_services(config: BodyworkConfig) -> bool:
    service_stages = [
        stage_name
        for stage_name, stage in config.stages.items()
        if isinstance(stage, ServiceStageConfig) and stage_name in config.project.DAG
    ]
    deploys_service_stages = True if service_stages else False
    return deploys_service_stages
```

tests/integration/test_k8s_with_cluster.py
AlexIoannides: I thought that might be the case - thanks.

tests/integration/test_k8s_with_cluster.py
Marios85: Yes because the namespace is derived from the config and we are using the same one (from Github). This was my point the other day about having a separate cluster for testing because we can now impact on live stuff we have running on our current cluster. 

When we have the secrets change done then we can just add a different namespace in the config locally and then we can avoid collisions with each other. This is fine for now as it's just us two.

tests/integration/test_k8s_with_cluster.py
Marios85: Nope because the namespace is static i.e. it is obtained from the config.YAML

AlexIoannides:
The docstring for `setup_namespace_with_service_accounts_and_roles` ought to mention that it prepares a namespace for running deployment jobs - i.e. it should no longer refer to stages, etc.

src/bodywork/workflow_execution.py
Marios85: My point exactly!  A whole method and repetition of code for something that is covered by one bool. I can't agree that this is a more optimal solution in this instance.

Marios85: Also it's just one bool in the whole method.

src/bodywork/workflow_execution.py
Marios85: It's not required but either way this is a more elegant solution using the inline if statement. I use them in C# so I don't know why I neglect them in Python 🤷

src/bodywork/k8s/workflow_jobs.py
```diff
@@ -97,6 +103,18 @@ def configure_workflow_job(
     return job
 
 
+def _create_job_name(project_repo_url: str, project_repo_branch: str) -> str:
```
AlexIoannides, line 106: Do we have a test for this (implicit or explicit)?

Marios85: I remember have an implicit one somewhere but  I can't remember where, I had to mock it.

Marios85: test_configure_workflow_job

AlexIoannides:
Just an aside/discussion point, but in the `test_workflow_and_service_management_end_to_end_from_cli`, am I correct in thinking that we don't _need_ to do `bodywork setup-namespace bodywork-test-project`, we could just do `kubectl create ns bodywork-test-project`? The bodywork version prepares a namespace for running workflows, whereas for these tests the workflow are still being run in the `bodywork-deployment-jobs` namespace? i.e. it's just being used as a convenience for the time-being?

Marios85:
It's just because we need to add the secrets to it. When we complete the secrets tickets and remove their association with namespaces nthen we can remove it as Bodywork will setup the namespace automatically. 

I'll add a note to the ticket to make sure I don't forget to do it.

src/bodywork/workflow_execution.py
AlexIoannides: Yeah, but you re-assign a value to the dummy variable, every time you hit that part of the code (i.e. for every step), which is far from elegant. At least `config_deploys_services` is just one-pass of the config file - it's just one list comprehension.

Marios85: Assigning to the bool again isn't optimal but it is far more efficient than iterating through the entire list again as assigning to a bool is a very cheap operation.  Also it takes fewer lines of code. 

What annoys me is that Python misses C# LINQ functions, specifically to use an Any() function with my parameters. Thinking about how that works I think the following is a compromise whilst still being fairly efficient:

`def _deploys_services(config: BodyworkConfig) -> bool:
    for stage_name, stage in config.stages.items():
        if isinstance(stage, ServiceStageConfig) and stage_name in config.project.DAG:
            return True
    return False`

Marios85: Assigning to the bool again isn't optimal but it is far more efficient than iterating through the entire list again as assigning to a bool is a very cheap operation. Also it takes fewer lines of code.

What annoys me is that Python misses C# LINQ functions, specifically to use an Any() function with my parameters. Thinking about how that works I think the following is a compromise whilst still being fairly efficient:

```python
def _deploys_services(config: BodyworkConfig) -> bool:
    for stage_name, stage in config.stages.items():
        if isinstance(stage, ServiceStageConfig) and stage_name in config.project.DAG:
            return True
    return False
    ```

Comment (AlexIoannides) [reply src/bodywork/workflow_execution.py:100]: There is an `any`:

```python
def config_deploys_services(config: BodyworkConfig) -> bool:
    return any(True for stage_name, stage in config.stages.items()
        if isinstance(stage, ServiceStageConfig) and stage_name in config.project.DAG
    )
```

?

Marios85: We have a winner! I didn't think we could use it like that from the docs but that's how I wanted to use it although the syntax is different.

AlexIoannides: 💥

src/bodywork/cli/workflow_jobs.py
```diff
@@ -75,7 +74,7 @@ def delete_workflow_job_in_namespace(namespace: str, job_name: str) -> None:
 
 def create_workflow_cronjob(
```
AlexIoannides, line 75: Is it worth keeping `namespace` as an arg here, given that it's an arg in every other function in this module and then setting it to `BODYWORK_DEPLOYMENT_JOBS_NAMESPACE` in `cli.py`?

src/bodywork/workflow_execution.py
```diff
@@ -183,25 +181,36 @@ def _setup_namespace(config) -> str:
...
-            k8s.setup_job_and_deployment_service_account(namespace)
+            k8s.setup_stages_service_account(namespace)
         return namespace
     except ApiException as e:
         raise BodyworkNamespaceError(
             f"Unable to setup namespace: {namespace} - {e}"
         ) from e
 
 
+def config_deploys_services(config: BodyworkConfig) -> bool:
```
AlexIoannides, line 201: Maybe `workflow_deploys_services` is better?

AlexIoannides: Maybe a simple test for this, too?

AlexIoannides (approved):
I have 3 small comments, but those aside, good-to-go!

It's an immense piece of work!

src/bodywork/cli/workflow_jobs.py
Marios85: Yes, I it's more consistent.

src/bodywork/workflow_execution.py
Marios85: The name was your suggestion. I prefer `workflow_deploys_services `. 

This is implicitly tested in `workflow_execution.py` in the last two tests.

AlexIoannides: Yes, that's right - you test the tear-down!

PR #118, opened 2021-07-26
116 make secrets namespace agnostic

### LINKED ISSUE #116: Make Secrets Namespace Agnostic
_As a Machine Learning Engineer I would like to use the same set of secrets across multiple workflows/deployments and not deal with namespaces._ 

**Tasks**

Currently Secrets are created for each individual namespace and instead they should be arranged into groups and be namespace agnostic.

- For the `bodywork secret` commands, replace namespace argument with a name that represents the name of the secrets group this secret is in e.g. --group. 
- Create all secrets in the `bodywork-deployment-jobs` namespace. 
- Prefix the name of the secret with the name arg provided or similar so that they can be retrieved according to the group name that they belong to e.g. 'Prod' group secret would be 'Prod-SSHKey'.
- Add `secrets_group` item to project section of bodywork.yaml
- Ensure the relevant secrets are retrieved and added to the workflow container/workflow if `secrets_group` is specified in the config. 

N.B. Remember to remove the namespace setup and amend the Secret creation in `test_workflow_and_service_management_end_to_end_from_cli`

Marios85:
This PR closes #116 . 

Changes mostly follow those described in #116. 

When replicating the secrets to the target namespace the workflow will run in, all the secrets that belong to the secrets group specified will be copied. This occurs once per workflow before any stages are executed. 

Created a separate Integration test file for Secrets as another small step towards splitting them out and cleaning up our tests.


tests/integration/conftest.py
```diff
@@ -150,3 +152,43 @@ def delete_namespace():
...
+@fixture(scope="function")
+def add_secrets(request: FixtureRequest) -> None:
+    run(
+        [
+            "kubectl",
+            "create",
+            "secret",
+            "generic",
+            f"--namespace={BODYWORK_DEPLOYMENT_JOBS_NAMESPACE}",
+            "test-bodywork-test-project-credentials",
```
AlexIoannides, line 166: Would `bodywork-test-project-credentials` not work as a name?

tests/integration/test_k8s_with_cluster.py
```diff
@@ -43,38 +38,12 @@
...
-                "USERNAME=alex",
-                "PASSWORD=alex123",
-            ],
-            encoding="utf-8",
-            capture_output=True,
-        )
-        assert process_one.stdout is not None
-        assert process_one.returncode == 0
-
         process_two = run(
```
AlexIoannides, line 47: This is now `process_one` - these night need a quick and tedious renaming.

AlexIoannides (approved):
A couple of minor comments, but otherwise this is right on-the-money 😎

tests/integration/conftest.py
AlexIoannides: I understand now - maybe something like `secretsgroup-bodywork-test-project-credentials`?

AlexIoannides:
This is not related to your PR, but while reviewing `workflow_execution.py` I've noticed that we might be missing an `except` in this try-catch block:

image

src/bodywork/workflow_execution.py
```diff
@@ -97,6 +97,12 @@ def run_workflow(
         env_vars = k8s.create_k8s_environment_variables(
             [(GIT_COMMIT_HASH_K8S_ENV_VAR, get_git_commit_hash(cloned_repo_dir))]
         )
+        if config.project.secrets_group:
+            _log.info(
+                f"Replicating {config.project.secrets_group} secrets in "
+                f"namespace={namespace}"
+            )
+            k8s.replicate_secrets_in_namespace(namespace, config.project.secrets_group)
```
AlexIoannides, line 105: Should we explicitly handle the potential exceptions from this block, so that the log message are clear about what has gone wrong? Otherwise, we could be feeding the user a k8s API error payload, which is somewhat verbose.

AlexIoannides:
Need to copy the `GIT_SSH_...` environment variable into the target namespace!!!

src/bodywork/cli/secrets.py
```diff
@@ -56,55 +56,72 @@ def parse_cli_secrets_strings(key_value_strings: Iterable[str]) -> Dict[str, str
...
 
     :param namespace: Namespace in which to look for secrets.
+    :param group: Group the secrets to display belong to.
     :param secret: Display the available keys in just the secret with
         this name, defaults to None
     """
     if not k8s.namespace_exists(namespace):
         print(f"namespace={namespace} could not be found on k8s cluster")
         return None
+    if secret and group is None:
```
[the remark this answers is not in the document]
Marios85: No that should not be possible because the purpose of this `If` statement is to make sure that `secret` can only be looked up if `group` is not none therefore you won't be able to get to `_create_complete_secret_name`.

Also `if group is None` would prevent displaying all secrets regardless of group which happens when we don't specify any optional parameters.

src/bodywork/k8s/secrets.py
```diff
@@ -121,16 +162,19 @@ def delete_secret(namespace: str, name: str) -> None:
...
+    :param group: Group of secrets to list.
+
     """
     secrets = k8s.CoreV1Api().list_namespaced_secret(
         namespace=namespace,
+        label_selector=f"{SECRET_GROUP_LABEL}={group}",
     )
     secret_data_base64 = {
-        s.metadata.name: s.string_data if s.string_data else s.data
+        f"{s.metadata.name}": s.string_data if s.string_data else s.data
```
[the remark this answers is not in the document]
Marios85: Actually there's no need even for that, we can revert it back to `s.metadata.name`.

tests/integration/conftest.py
Marios85: I think test_secrets would be a better name that's also more descriptive so you'll know that these secrets are to do with tests when looking at the Bodywork namespace and also to avoid secrets_group=secretsgroup in the test project YAML which seems a bit off.

tests/integration/test_k8s_with_cluster.py
Marios85: Cheers I must have forgotten to amend this one.

src/bodywork/workflow_execution.py
Marios85: I agree, a one liner explaining what went wrong would make it easier for the user to understand at what point things failed. I'll extract the replicating of the secrets into a separate method as well to try and keep things neat and succinct in the `run_workflow` method.

src/bodywork/cli/secrets.py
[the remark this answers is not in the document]
AlexIoannides: I'm confused. If `secret = "foo"` and `group = None`, then we enter the `try` block and run,

```python
_create_complete_secret_name(cast(str, None), "foo")
```

And I don't think the `KeyError` will catch this.

What am I missing?

tests/integration/conftest.py
AlexIoannides: Agreed.

src/bodywork/cli/secrets.py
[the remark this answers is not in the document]
Marios85: If `secret = "foo"` and `group = None` then it satisfies the following condition and the method exits with None.
    if secret and group is None:
        print("please specify which secrets group the secret belongs to.")
        return None
`If secret` means if secret has a value not if both secret and group are None.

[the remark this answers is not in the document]
AlexIoannides: Aaaahhh - yes. It's more like,

```python
if secret and (group is None):
    # ...
```

I was reading it literally. I am so foggy today...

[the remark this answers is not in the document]
Marios85: Me too, I did a "double take" when looking at this as well.

Marios85:
Good spot with the missing catch block. I don't know how that went under the radar for so long. 

The SSH key is now passed through as an environment variable to the target namespace. We should have done this earlier instead of expecting the user to create a secret for it.

AlexIoannides:
I'm not sure what you mean by,

> "_The SSH key is now passed through as an environment variable to the target namespace. We should have done this earlier instead of expecting the user to create a secret for it._"

PR #125, opened 2021-08-15
121 display services in all namespaces cli command

### LINKED ISSUE #121: Display Services in all namespaces - CLI command
"As an ML Engineer I would like to see all the deployed services (on the cluster)".

Extend the existing 'Services' CLI command with the option to view all the services on the cluster which is equivalent to `kubectl get services --all-namespaces`

Bodywork CLI command will be `bodywork service display`

Marios85:
This PR closes #121 . 

Leveraged the existing `bodywork service display` command so that if it is used without a projectname/namespace arg then all deployed services on the cluster are returned. 

One change of significance is in the method `k8s\service_deployments\list_service_stage_deployments`. I replaced the deployment_info generator with a conventional 'for loop' because the API was being called 3 times unnecessarily when it can be called once as per the refactored code. 

Also added the ability to search for a specific service in all or a specific namespace.


src/bodywork/cli/cli.py
```diff
@@ -474,9 +474,12 @@ def service(args: Namespace) -> None:
     elif command == "delete":
         load_kubernetes_config()
         delete_service_deployment_in_namespace(namespace, name)
+    elif name is not None:
```
AlexIoannides, line 477: Could you get rid of this `elif` block by specifying `None` as a default value in,

```python
service_cmd_parser.add_argument(
     "--namespace",
     "--ns",
     required=False,
     type=str,
     help="Kubernetes namespace to operate in.",
 )
```

?

src/bodywork/k8s/service_deployments.py
```diff
@@ -402,16 +404,15 @@ def list_service_stage_deployments(namespace: str) -> Dict[str, Dict[str, str]]:
             "git_url": deployment.spec.template.spec.containers[0].args[0],
             "git_branch": deployment.spec.template.spec.containers[0].args[1],
             "has_ingress": (
-                "true" if has_ingress(namespace, deployment.metadata.name) else "false"
+                has_ingress(deployment.metadata.namespace, deployment.metadata.name)
```
AlexIoannides, line 407: This won't return a string, it'll return a bool - think we should leave this one as-is.

tests/unit_and_functional/test_cli_service_deployments.py
```diff
@@ -24,7 +24,7 @@
 
 from bodywork.cli.service_deployments import (
     delete_service_deployment_in_namespace,
-    display_service_deployments_in_namespace,
+    display_service_deployments,
```
AlexIoannides, line 27: Do we need to add a test for the no namespace condition?

src/bodywork/k8s/service_deployments.py
AlexIoannides: I'm amazed MyPy didn't flag this, because the return type is `Dict[str, str]`.

tests/unit_and_functional/test_k8s_service_deployments.py
```diff
@@ -471,17 +472,82 @@ def test_list_service_stage_deployments_returns_service_stage_info(
...
+
+@patch("kubernetes.client.AppsV1Api")
+def test_list_service_stage_deployments_returns_all_services_on_cluster(
+    mock_k8s_apps_api: MagicMock,
+    service_stage_deployment_object: kubernetes.client.V1Deployment,
+):
+    service_namespace = service_stage_deployment_object.metadata.namespace
+    service_name = service_stage_deployment_object.metadata.name
+
+    service_stage_deployment_object.status = kubernetes.client.V1DeploymentStatus(
```
AlexIoannides, line 499: An anally retentive voice in my head is telling me that this should be named `service_stage_deployment_object1`...

tests/unit_and_functional/test_k8s_service_deployments.py
```diff
@@ -471,17 +472,82 @@ def test_list_service_stage_deployments_returns_service_stage_info(
...
+        available_replicas=1, unavailable_replicas=None
+    )
+
+    service_stage_deployment_object2 = copy.deepcopy(service_stage_deployment_object)
+    service_stage_deployment_object2.metadata.name = "deployment two"
```
AlexIoannides, line 504: `deployment-two` - I think whitespace is an invalid k8s names. Not that is matters, but...

AlexIoannides (changes requested):
A few very minor comments - otherwise all good 😎

src/bodywork/cli/cli.py
Marios85: I think `None` will be the default if no default is specified but we want to make sure we filter on empty strings as well which I believe would occur if you did `--name` but forgot to specify a value. 

However this `if` statement should be just `elif name` so i'll fix that. Also I can remove some repetition in the whole block so I am going to do a little bit of refactoring too.

src/bodywork/k8s/service_deployments.py
Marios85: That's because a bool will automatically convert to a string because the bool type has a "to_string()" function (or whatever the Python equivalent of that is) therefore this will return a string with either "True" or "False".

tests/unit_and_functional/test_cli_service_deployments.py
Marios85: Yes we should because it is effectively a new use case i.e. Display all service deployments, even though we do test this in `k8s.service_deployments`.

tests/unit_and_functional/test_k8s_service_deployments.py
Marios85: Changed to silence the voice 🙂

tests/unit_and_functional/test_k8s_service_deployments.py
Marios85: Done!

tests/unit_and_functional/test_k8s_service_deployments.py
Marios85: Reverting this change, it's because that's the name of the fixture that returns the object up top therefore it would be weird to rename this with the 1 at the end (and in all the other tests that use it).

src/bodywork/cli/cli.py
AlexIoannides: You're correct - `None` is the default (just checked). However, you'll get an error if you forgot to specify a value for `--name`, so no need to filter on empty strings unless someone explicitly passes `""`.

src/bodywork/k8s/service_deployments.py
AlexIoannides: No it won't - there's no such thing in Python. At most, `""` equates to `False` and `"anystring"` equates to `True`, but that's it.

tests/unit_and_functional/test_cli_service_deployments.py
```diff
@@ -27,6 +27,20 @@
     display_service_deployments,
 )
 
+Test_ServiceStageDeployment = {
```
AlexIoannides, line 30: Should probably be a fixture, or at the very least use camelcase for name if it isn't.

src/bodywork/k8s/service_deployments.py
Marios85: There seems to be some native translation of bools to strings because this works. Try typing the following into a Python Console and look at the result:

str(bool(True))
str(bool(False))

It will return strings with `True` and `False` respectively. This is effectively what is happening in this code snippet. `has_ingress()` returns a bool not a string.

src/bodywork/cli/cli.py
Marios85: --name is not a required parameter i.e. `required=false` therefore an error won't be thrown if it's not specified. Either way it's best to leave the code defensive and cover the explicit `""` scenario

tests/unit_and_functional/test_cli_service_deployments.py
Marios85: Good spot, it was meant be camelcase but I put an underscore there by mistake out of force of habit from writing tests. 

I'll make it a fixture just in case we manipulate this object in future tests.

src/bodywork/cli/cli.py
AlexIoannides: If you don't specify `--name` does it equal `""` or `None`?

src/bodywork/k8s/service_deployments.py
AlexIoannides: But you have to force the conversion - e.g. by calling `str(obj)`. It won't happen automatically otherwise - the Python interpreter completely ignores type hints.

tests/unit_and_functional/test_cli_service_deployments.py
AlexIoannides: Nice one 👍

src/bodywork/k8s/service_deployments.py
Marios85: Change the code from my previous comment with `print(bool(True))` or do `f"{bool(True)}"` and you will get the same result. It returns the True/False string value that we require. 

I see no value add by explicitly declaring the strings in a conditional statement and it's not as simple. It's effectively a custom implementation of casting a bool to a string.

src/bodywork/cli/cli.py
Marios85: This is obsolete now, I forgot to mention in the comments (I think I just did on Slack) that I implemented the feature so that we can search for an individual service so now we use --name and it is the same method that is called with optional parameters so this is now just an `else`.

src/bodywork/k8s/service_deployments.py
AlexIoannides: Got it - yes. And this'll be replaced by a DTO at some point, in which case we want the `bool` anyway.

Should probably change `Dict[str, str]` -> `Dict[str, Union[str, bool]]` or `Dict[str, Any]` to that it's accurate.

src/bodywork/cli/cli.py
AlexIoannides: 💥

src/bodywork/k8s/service_deployments.py
Marios85: I've changed it to `Dict[str, Any]`, we also have `int` types in there as well e.g. available_replicas.

PR #126, opened 2021-08-18
123 update secret command

### LINKED ISSUE #123: Update Secret Command
"As an ML Engineer I would like to update an existing Secret"

Extend the Secrets CLI command with a sub command to update a secret i.e. `bodywork secret update --ns group-name --name mysecret --data USERNAME=marios PASSWORD=xyz`

Marios85:
This PR closes #123 . This PR adds the ability to update a secret via the cli. 

In addition to this the PR contains a refactor of the display all Secrets command (including a bug fix) so that a (Bodywork) Secrets object is returned. This allows us to return additional fields such as Group and leaves it extensible for the future.

### DIFF
--- a/src/bodywork/cli/cli.py
+++ b/src/bodywork/cli/cli.py
@@ -491,10 +491,14 @@ def secret(args: Namespace) -> None:
     group = args.group
     name = args.name
     key_value_strings = args.data
-    if (command == "create" or command == "delete" or command == "update") and name == "":
+    if (
+        command == "create" or command == "delete" or command == "update"
+    ) and name == "":
         print("please specify the name of the secret")
         sys.exit(1)
-    if (command == "create" or command == "delete" or command == "update") and group == "":
+    if (
+        command == "create" or command == "delete" or command == "update"
+    ) and group == "":
         print("please specify the secret group the secret belongs to")
         sys.exit(1)
     elif (command == "create" or command == "update") and key_value_strings == []:
@@ -515,7 +519,9 @@ def secret(args: Namespace) -> None:
                 BODYWORK_DEPLOYMENT_JOBS_NAMESPACE, group, name, var_names_and_values
             )
         else:
-            update_secret(BODYWORK_DEPLOYMENT_JOBS_NAMESPACE, group, name, var_names_and_values)
+            update_secret(
+                BODYWORK_DEPLOYMENT_JOBS_NAMESPACE, group, name, var_names_and_values
+            )
     elif command == "delete":
         load_kubernetes_config()
         delete_secret(BODYWORK_DEPLOYMENT_JOBS_NAMESPACE, group, name)
--- a/src/bodywork/cli/cli.py
+++ b/src/bodywork/cli/cli.py
@@ -49,6 +49,7 @@ from .secrets import (
     delete_secret,
     display_secrets,
     parse_cli_secrets_strings,
+    update_secret,
 )
 from .setup_namespace import (
     is_namespace_available_for_bodywork,
@@ -199,7 +200,7 @@ def cli() -> None:
     secret_cmd_parser.add_argument(
         "command",
         type=str,
-        choices=["create", "delete", "display"],
+        choices=["create", "delete", "update", "display"],
         help="Secrets action to perform.",
     )
     secret_cmd_parser.add_argument(
@@ -490,16 +491,16 @@ def secret(args: Namespace) -> None:
     group = args.group
     name = args.name
     key_value_strings = args.data
-    if (command == "create" or command == "delete") and name == "":
-        print("please specify a name for the secret you want to create")
+    if (command == "create" or command == "delete" or command == "update") and name == "":
+        print("please specify the name of the secret")
         sys.exit(1)
-    if (command == "create" or command == "delete") and group == "":
+    if (command == "create" or command == "delete" or command == "update") and group == "":
         print("please specify the secret group the secret belongs to")
         sys.exit(1)
-    elif command == "create" and key_value_strings == []:
-        print("please specify keys and values for the secret you want to create")
+    elif (command == "create" or command == "update") and key_value_strings == []:
+        print("please specify keys and values for the secret you want to create/update")
         sys.exit(1)
-    elif command == "create":
+    elif command == "create" or command == "update":
         try:
             var_names_and_values = parse_cli_secrets_strings(key_value_strings)
         except ValueError:
@@ -509,9 +510,12 @@ def secret(args: Namespace) -> None:
             )
             sys.exit(1)
         load_kubernetes_config()
-        create_secret(
-            BODYWORK_DEPLOYMENT_JOBS_NAMESPACE, group, name, var_names_and_values
-        )
+        if command == "create":
+            create_secret(
+                BODYWORK_DEPLOYMENT_JOBS_NAMESPACE, group, name, var_names_and_values
+            )
+        else:
+            update_secret(BODYWORK_DEPLOYMENT_JOBS_NAMESPACE, group, name, var_names_and_values)
     elif command == "delete":
         load_kubernetes_config()
         delete_secret(BODYWORK_DEPLOYMENT_JOBS_NAMESPACE, group, name)
--- a/src/bodywork/cli/cli.py
+++ b/src/bodywork/cli/cli.py
@@ -207,11 +207,10 @@ def cli() -> None:
         "--group",
         required=False,
         type=str,
-        default="",
         help="The secrets group this secret belongs in.",
     )
     secret_cmd_parser.add_argument(
-        "--name", type=str, default="", help="The name given to the Kubernetes secret."
+        "--name", type=str, help="The name given to the Kubernetes secret."
     )
     secret_cmd_parser.add_argument(
         "--data",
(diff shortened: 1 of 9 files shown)


src/bodywork/cli/cli.py
```diff
@@ -523,7 +532,7 @@ def secret(args: Namespace) -> None:
         display_secrets(
             BODYWORK_DEPLOYMENT_JOBS_NAMESPACE,
             group,
-            secret=name if name != "" else None,
+            name,
```
AlexIoannides, line 535: Would this prevent us from seeing all secrets in a group?

This would make it tricky to keep track of what secrets have been created.

src/bodywork/k8s/secrets.py
```diff
@@ -18,15 +18,24 @@
...
 
 from kubernetes import client as k8s
 
 from .utils import make_valid_k8s_name
 from ..constants import SECRET_GROUP_LABEL, BODYWORK_DEPLOYMENT_JOBS_NAMESPACE
 
 
+@dataclass
+class Secret:
+    def __init__(self, name: str, group: str, data: Dict[str, str]):
```
AlexIoannides, line 33: You don't need an `__init__` for a data class.

```python
@dataclass
 class Secret:
    self.name = name
    self.group = group
    self.data = data
```

Will work.

src/bodywork/k8s/secrets.py
```diff
@@ -162,25 +189,32 @@ def delete_secret(namespace: str, name: str) -> None:
     k8s.CoreV1Api().delete_namespaced_secret(namespace=namespace, name=name)
 
 
-def list_secrets(namespace: str, group: str = None) -> Dict[str, Dict[str, str]]:
+def list_secrets(namespace: str, group: str = None) -> Dict[str, Secret]:
```
AlexIoannides, line 192: Should be `group: Optional[str] = None` - I'm amazed that MyPy didn't complain about this. I hope I haven't badly configured the settings...

src/bodywork/k8s/secrets.py
```diff
@@ -162,25 +189,32 @@ def delete_secret(namespace: str, name: str) -> None:
...
+            s.metadata.labels[SECRET_GROUP_LABEL]
+            if s.metadata.labels and SECRET_GROUP_LABEL in s.metadata.labels
+            else None,
+            s.string_data if s.string_data else s.data,
+        )
+        for s in result.items
     }
-    return secret_data_decoded
+
+    for key, value in secrets.items():
```
AlexIoannides, line 217: Can I suggest that you roll this into the creation of the `secrets` dict or add the logic into the constructor for the `Secrets` class?

tests/unit_and_functional/test_cli_secrets.py
```diff
@@ -94,31 +116,32 @@ def test_delete_secrets_in_namespace(
...
 
 @patch("bodywork.cli.secrets.k8s")
-def test_display_secrets(
-    mock_k8s_module: MagicMock, capsys: CaptureFixture
-):
+def test_display_secrets(mock_k8s_module: MagicMock, capsys: CaptureFixture):
     mock_k8s_module.list_secrets.return_value = {
-        "PROD-test-credentials": {"USERNAME": "alex", "PASSWORD": "alex123"},
-        "DEV-more-test-credentials": {"FOO": "bar"},
+            "PROD-test-credentials": Secret("PROD-test-credentials", "PROD", {"USERNAME": "alex", "PASSWORD": "alex123"}),
```
AlexIoannides, line 128: I think you might need to run Black on this file.

tests/unit_and_functional/test_k8s_secrets.py
```diff
@@ -126,7 +146,7 @@ def test_list_secrets_in_namespace_returns_decoded_secret_data(
             items=[
                 kubernetes.client.V1Secret(
                     metadata=kubernetes.client.V1ObjectMeta(
-                        namespace="bodywork-dev", name="pytest-secret", labels="xyz"
+                        namespace="bodywork-dev", name="xyz-pytest-secret", labels={f"{SECRET_GROUP_LABEL}": "xyz"}
```
AlexIoannides, line 149: Need to run Black on this file to fix line length?

AlexIoannides (changes requested):
Just some minor comments, otherwise grade A (as always).

src/bodywork/k8s/secrets.py
Marios85: Good spot, that's weird it didn't get picked up. I had a quick look and our MyPy is using the version that should check this as this was allowed in older versions of the PEP rules.

Also MyPy looks configured correctly, my IDE was picking this up as optional without the explicit declaration as well.

src/bodywork/k8s/secrets.py
Marios85: I disagree here, this loop contains a line to decode to UTF8. You would end up restricting the creation of the Secrets Dataclass just for Secrets obtained directly from the API i.e. just from this method. 

A DTO/dataclass should not impose that restriction, the data needs to be able to flow the other way.

src/bodywork/k8s/secrets.py
AlexIoannides: Okay, so you can now have implicit optional types - see [here](https://mypy.readthedocs.io/en/stable/command_line.html#none-and-optional-handling).

I don't mind either way, but we should keep it uniform across the codebase. This might mean using `no_implicit_optional` in the MyPy config file?

src/bodywork/k8s/secrets.py
AlexIoannides: What I'm taking exception to here, is mutating the `Secrets` object (the `data` field) from the outside. This doesn't feel right to me. I'd rather there was a `data_decoded` field so this wasn't necessary. Have I understood this correctly?

src/bodywork/k8s/secrets.py
Marios85: My vote is for the implied version as it is less verbose and more natural for me as that is similar to how they are defined in C#. 

I agree they should be the same everywhere. I don't mind doing the change with a replace all but I would create a separate PR for doing this to keep things clean.

AlexIoannides: I agree with you - go for implied and then change them all in one dedicated PR.

src/bodywork/k8s/secrets.py
Marios85: I understand where you're coming from but I think it's unnecessary because when we create or update a new secret we do not need to encode it to be able to save/update, that is handled by Kubernetes so we only ever use plain text.  

This object is our simplified abstraction of a kubernetes Secret. I cannot see the value add of keeping the encoded data in the object in addition to the decoded data, I cannot see a potential use for that.

In the previous implementation we returned just the decoded data in the dictionary so there is no change in that respect either.

Maybe it's easier to discuss this one in person.

src/bodywork/k8s/secrets.py
Marios85: Cool, i'll do this once all the outstanding PR's have been merged.

src/bodywork/cli/cli.py
AlexIoannides: I reckon `name if name else None` ought to do it as the default is `None` as we've established elsewhere.

src/bodywork/k8s/secrets.py
AlexIoannides: Yeah, let's chat about this.

AlexIoannides: The other option is to decode elsewhere - i.e. wherever it needs to be decoded, e.g. in the CLI function for printing to stdout, etc.

src/bodywork/cli/cli.py
Marios85: I'm lost here, what you're suggesting will have no effect. `name` gets passed through whether it's `none` or not at the moment. Also currently it allows all secrets in a group to be listed by doing `display secret --group Prod`.

Am I missing something here?

src/bodywork/k8s/secrets.py
Marios85: It's now decoded in the constructor, we should have done this in the first place instead of looping through the values afterwards. 

Also the previous implementations would decode the string_data field if that was returned. This would result in an error if that ever happened because the string_data field is just for decoded data i.e. plain text.

src/bodywork/cli/cli.py
AlexIoannides: It's me that's the problem here. I need to stop reviewing code on trains - need a big screen to do it properly.

AlexIoannides: It's good - sorry!

src/bodywork/k8s/secrets.py
AlexIoannides: Nice one - looks good.

Have you blacked that file? I'm curious if that block is kosher or not.

src/bodywork/cli/cli.py
Marios85: No worries! I'd struggle to do it on the train as well but for me it's more because there's too many distractions.

src/bodywork/k8s/secrets.py
Marios85: Yeah it's Black that formatted it that way. It does look a bit weird to me as well, I don't like how the `else None` is on a separate line when there is space to fit it on the same line.

PR #127, opened 2021-08-23
122 update cronjob command

### LINKED ISSUE #122: Update Cronjob Command
"As an ML Engineer I would like to update an existing Cronjob"

Create the CLI command option to update an existing cronjob i.e. `bodywork cronjob update --ns project-name --name cronjob-name --schedule * * * 5`

Marios85:
This PR closes #122 . 

The update cronjob command allows a cronjob to be updated with all the currently available parameters that can specified when creating a cronjob. The only constraint imposed is that when updating either the git branch or URL, both must be specified together. 

If the cronjob schedule is not defined then the existing schedule must be retrieved because this a mandatory field when creating a cronjob.

### DIFF
--- a/src/bodywork/cli/workflow_jobs.py
+++ b/src/bodywork/cli/workflow_jobs.py
@@ -19,6 +19,7 @@ This module contains functions for managing Kubernetes workflow jobs.
 They are targeted for use via the CLI.
 """
 import re
+from typing import Optional
 
 from .. import k8s
 
@@ -45,7 +46,7 @@ def create_workflow_job_in_namespace(
     if not k8s.namespace_exists(namespace):
         print(f"namespace={namespace} could not be found on k8s cluster")
         return None
-    if job_name and _is_existing_workflow_job(namespace, job_name):
+    if _is_existing_workflow_job(namespace, job_name):
         print(f"workflow job={job_name} already exists in namespace={namespace}")
         return None
     configured_job = k8s.configure_workflow_job(
@@ -121,11 +122,11 @@ def create_workflow_cronjob(
 def update_workflow_cronjob_in_namespace(
     namespace: str,
     job_name: str,
-    schedule: str = None,
-    project_repo_url: str = None,
-    project_repo_branch: str = None,
-    retries: int = None,
-    workflow_job_history_limit: int = None,
+    schedule: Optional[str] = None,
+    project_repo_url: Optional[str] = None,
+    project_repo_branch: Optional[str] = None,
+    retries: Optional[int] = None,
+    workflow_job_history_limit: Optional[int] = None,
 ) -> None:
     """Update a new cronjob within a namespace.
 
--- a/src/bodywork/cli/workflow_jobs.py
+++ b/src/bodywork/cli/workflow_jobs.py
@@ -49,7 +49,7 @@ def create_workflow_job_in_namespace(
         print(f"workflow job={job_name} already exists in namespace={namespace}")
         return None
     configured_job = k8s.configure_workflow_job(
-        namespace, project_repo_url, project_repo_branch, retries
+        namespace, project_repo_url, project_repo_branch, retries, job_name=job_name
     )
     k8s.create_workflow_job(configured_job)
     print(f"workflow job={job_name} created in namespace={namespace}")
--- a/src/bodywork/cli/workflow_jobs.py
+++ b/src/bodywork/cli/workflow_jobs.py
@@ -25,9 +25,9 @@ from .. import k8s
 
 def create_workflow_job_in_namespace(
     namespace: str,
+    job_name: str,
     project_repo_url: str,
     project_repo_branch: str = "master",
-    job_name: str = None,
     retries: int = 2,
 ) -> None:
     """Create a new workflow job within a namespace.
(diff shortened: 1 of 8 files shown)


src/bodywork/cli/workflow_jobs.py
```diff
@@ -45,11 +45,11 @@ def create_workflow_job_in_namespace(
     if not k8s.namespace_exists(namespace):
         print(f"namespace={namespace} could not be found on k8s cluster")
         return None
-    if _is_existing_workflow_job(namespace, job_name):
+    if job_name and _is_existing_workflow_job(namespace, job_name):
```
AlexIoannides, line 48: I'm not sure you need the `and` if `job_name` cannot be `None` - i.e. if it isn't `Optional[str]`.

src/bodywork/cli/workflow_jobs.py
```diff
@@ -119,9 +115,53 @@ def create_workflow_cronjob(
...
     k8s.create_workflow_cronjob(configured_job)
-    print(
-        f"workflow cronjob={job_name} created in {namespace} namespace."   # noqa
+    print(f"workflow cronjob={job_name} created in {namespace} namespace.")  # noqa
+
+
+def update_workflow_cronjob_in_namespace(
+    namespace: str,
+    job_name: str,
+    schedule: str = None,
```
AlexIoannides, line 124: Should be `Optional[str] = None `.

src/bodywork/cli/workflow_jobs.py
```diff
@@ -119,9 +115,53 @@ def create_workflow_cronjob(
...
+    project_repo_url: str = None,
```
AlexIoannides, line 125: Should also be `Optional[str]` and the same goes for the lines below.

src/bodywork/k8s/workflow_jobs.py
```diff
@@ -185,6 +183,70 @@ def create_workflow_cronjob(cron_job: k8s.V1Job) -> None:
     )
 
 
+def update_workflow_cronjob(
+    namespace: str,
+    name: str,
+    schedule: str = None,
```
AlexIoannides, line 189: `Optional[str] = None` + check the args below.

Might be worth doing a `ctrl+F` on the codebase, just in case I've missed any. I can't believe that MyPy hasn't complained.

src/bodywork/k8s/workflow_jobs.py
```diff
@@ -185,6 +183,70 @@ def create_workflow_cronjob(cron_job: k8s.V1Job) -> None:
...
+        )
+        job_template = k8s.V1beta1JobTemplateSpec(spec=job.spec)
+    else:
+        job_template = k8s.V1beta1JobTemplateSpec()
+
+    cronjob = k8s.V1beta1CronJob(
+        spec=k8s.V1beta1CronJobSpec(
+            job_template=job_template,
+            schedule=schedule,
+            successful_jobs_history_limit=successful_jobs_history_limit,
```
AlexIoannides, line 242: Should we not need to retrieve `successful_jobs_history_limit` and `failed_jobs_history_limit` if they're `None`?

tests/unit_and_functional/test_k8s_workflow_jobs.py
```diff
@@ -22,7 +22,7 @@
 from datetime import datetime
 from unittest.mock import MagicMock, patch
 
-import kubernetes
+import kubernetes.client as k8s
```
AlexIoannides, line 25: Should probably do this across all the tests for the sake of uniformity?

AlexIoannides: Although I must admit, that calling it `k8s` confuses it with our internal `k8s` sub-module, which is why I kept the verbose version. It might me better as `k8s_client`?

AlexIoannides (changes requested):
Various bits and pieces. Need to fix the recurring lack of `Optional[xx] = None` etc.

src/bodywork/cli/workflow_jobs.py
Marios85: I have no idea why I put that there.

src/bodywork/k8s/workflow_jobs.py
Marios85: This is the update/patch method, if they are `None` then they are not updated. If we wrote a conditional  to omit these parameters if they are `None` then it would have no effect and therefore it's not necessary.

tests/unit_and_functional/test_k8s_workflow_jobs.py
Marios85: Yeah I agree, it could easily be confused when flipping between test files from Cli,K8s etc. I've gone with your suggestion.

src/bodywork/k8s/workflow_jobs.py
AlexIoannides: By mistake - I misunderstood how the `patch` methods work.

tests/unit_and_functional/test_k8s_workflow_jobs.py
AlexIoannides: I think you may have forgotten to do this (according to what I'm seeing at this moment in time)?

AlexIoannides: Ignore this - found the right button on GitHub...

Marios85: That's got me a few times as well 🙂

src/bodywork/k8s/workflow_jobs.py
Marios85: It took me a while to extract that information from the docs

PR #148, opened 2021-09-28
140 remove workflow command

### LINKED ISSUE #140: Remove Workflow Command
The `deployment create` command is equivalent to the `workflow` command therefore the `workflow` command should be removed to keep things clean and simple. 

This will involve updating the `bodywork` command used to run the workflows in `bodywork.k8s.workflow_jobs.configure_workflow_job`.

Marios85:
This PR closes #140 . 

The workflow command has been fully replaced with the `deployment create` command.

### DIFF
--- a/src/bodywork/cli/cli.py
+++ b/src/bodywork/cli/cli.py
@@ -230,19 +230,6 @@ def cli() -> None:
         "stage_name", type=str, help="The Bodywork project stage to execute."
     )
 
-    # workflow interface
-    workflow_cmd_parser = cli_arg_subparser.add_parser("workflow")
-    workflow_cmd_parser.set_defaults(func=workflow)
-    workflow_cmd_parser.add_argument("git_url", type=str, help="Bodywork project URL.")
-    workflow_cmd_parser.add_argument(
-        "git_branch", type=str, help="Bodywork project Git repo branch."
-    )
-    workflow_cmd_parser.add_argument(
-        "--bodywork-docker-image",
-        type=str,
-        help="Bodywork Docker image to use - must exist on Bodywork DockerHub repo.",
-    )
-
     # setup-namespace interface
     setup_namespace_cmd_parser = cli_arg_subparser.add_parser("setup-namespace")
     setup_namespace_cmd_parser.set_defaults(func=setup_namespace)
@@ -358,16 +345,12 @@ def deployment(args: Namespace) -> None:
         print_warn("Please specify --name for the deployment job.")
         sys.exit(1)
     if command == "create":
+        load_kubernetes_config()
         if not async_workflow:
-            pass_through_args = Namespace(
-                git_url=git_url,
-                git_branch=git_branch,
-                bodywork_docker_image=image,
-            )
             print_info("Using local workflow controller - retries inactive.")
-            workflow(pass_through_args)
+            run_workflow(git_url, git_branch, docker_image_override=image)
         else:
-            load_kubernetes_config()
+            print_info("Using async workflow controller.")
             if not is_namespace_available_for_bodywork(
                 BODYWORK_DEPLOYMENT_JOBS_NAMESPACE
             ):
@@ -545,23 +528,6 @@ def stage(args: Namespace) -> None:
         run_stage(stage_name, repo_url, repo_branch)
         sys.exit(0)
     except Exception:
-        sys.exit(1)
-
-
-@handle_k8s_exceptions
-def workflow(args: Namespace) -> None:
-    """Workflow execution handler
-
-    :param args: Arguments passed to the workflow command from the CLI.
-    """
-    try:
-        repo_url = args.git_url
-        repo_branch = args.git_branch
-        docker_image = args.bodywork_docker_image
-        load_kubernetes_config()
-        run_workflow(repo_url, repo_branch, docker_image_override=docker_image)
-        sys.exit(0)
-    except BodyworkWorkflowExecutionError:
         sys.exit(1)
 
 
--- a/src/bodywork/cli/cli.py
+++ b/src/bodywork/cli/cli.py
@@ -58,7 +58,6 @@ from ..exceptions import (
     BodyworkConfigValidationError,
     BodyworkConfigMissingSectionError,
     BodyworkConfigParsingError,
-    BodyworkWorkflowExecutionError,
 )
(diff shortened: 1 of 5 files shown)


src/bodywork/cli/cli.py
```diff
@@ -358,16 +344,12 @@ def deployment(args: Namespace) -> None:
...
-                git_url=git_url,
-                git_branch=git_branch,
-                bodywork_docker_image=image,
-            )
             print_info("Using local workflow controller - retries inactive.")
-            workflow(pass_through_args)
+            run_workflow(git_url, git_branch, docker_image_override=image)
         else:
-            load_kubernetes_config()
+            print_info("Using async workflow controller.")
```
AlexIoannides, line 352: `"Using asynchronous workflow controller."`

I don't think we need to shorten it.

src/bodywork/cli/cli.py
```diff
@@ -358,16 +344,12 @@ def deployment(args: Namespace) -> None:
...
+        load_kubernetes_config()
         if not async_workflow:
-            pass_through_args = Namespace(
-                git_url=git_url,
-                git_branch=git_branch,
-                bodywork_docker_image=image,
-            )
             print_info("Using local workflow controller - retries inactive.")
-            workflow(pass_through_args)
+            run_workflow(git_url, git_branch, docker_image_override=image)
```
AlexIoannides, line 350: We need to run this in a `try ... except` block and look for `BodyworkWorkflowExecutionError` just like we did for the original `workflow` function.

src/bodywork/cli/cli.py
```diff
@@ -548,23 +530,6 @@ def stage(args: Namespace) -> None:
...
-    :param args: Arguments passed to the workflow command from the CLI.
-    """
-    try:
-        repo_url = args.git_url
-        repo_branch = args.git_branch
-        docker_image = args.bodywork_docker_image
-        load_kubernetes_config()
-        run_workflow(repo_url, repo_branch, docker_image_override=docker_image)
-        sys.exit(0)
-    except BodyworkWorkflowExecutionError:
```
AlexIoannides, line 564: I think we need to retain this block.

tests/unit_and_functional/test_k8s_workflow_jobs.py
```diff
@@ -161,7 +180,15 @@ def test_configure_workflow_cronjob():
     assert cronjob_definition.spec.job_template.spec.backoff_limit == 2
     assert cronjob_definition.spec.job_template.spec.template.spec.containers[
         0
-    ].args == ["bodywork-ml/bodywork-test-project", "dev"]
+    ].command == [
```
AlexIoannides, line 183: Man alive! This is ugly - Black?

src/bodywork/k8s/workflow_jobs.py
```diff
@@ -72,8 +72,15 @@ def configure_workflow_job(
         image=image,
         image_pull_policy="Always",
         env=vcs_env_vars,
-        command=["bodywork", "workflow"],
-        args=[project_repo_url, project_repo_branch],
+        command=[
```
AlexIoannides, line 75: The command is `bodywork deployment create`, but the rest are args, no?

There is a reason why you should split things in this way, but I'm failing to remember my Docker 101, etc.

AlexIoannides (changes requested):
Just some minor comments, but otherwise good.

src/bodywork/cli/cli.py
Marios85: Should this not be done for all commands or was the intention to suppress any workflow errors being spat out again to the CLI?

Marios85: I can't remember

src/bodywork/cli/cli.py
Marios85: Same as your comment above about the Try and Except block, no?

tests/unit_and_functional/test_k8s_workflow_jobs.py
Marios85: Yeah that's Black

src/bodywork/k8s/workflow_jobs.py
Marios85: Hmmm, I thought we couldn't split it this way because they need the flags but then it might work just adding the flags to the args. 

I'd need to try it out, if it does work then that's probably better, I don't know why I didn't try that out first. I shouldn't code late evening.

src/bodywork/cli/cli.py
AlexIoannides: No exceptions are meant to get printed to the CLI - we want everything to fail gracefully, where possible.

We also want to ensure that the command exists with non-zero exit code, as this may be used in CICD pipelines, etc.

src/bodywork/cli/cli.py
AlexIoannides: Yes - all a bit stream-of-consciousness at the moment...

tests/unit_and_functional/test_k8s_workflow_jobs.py
AlexIoannides: It is what it is...

src/bodywork/k8s/workflow_jobs.py
AlexIoannides: Hit the k8s book in the morning, or perhaps better might be the official [Dockerfile best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/).

AlexIoannides: And yeah, you shouldn't be coding late at night - that's the sort of stupid sh*t that I do...

src/bodywork/cli/cli.py
Marios85: The non-zero aspect is fine but that means we should wrap all of the commands in a try/catch block.

AlexIoannides: We do for k8s API exceptions, by virtue of the `handle_k8s_exceptions` decorator. All of the other CLI functions deliberately don't throw exceptions - they handle them internally.

Marios85: ahhhh I forgot about the decorator function we had on everything else.

PR #149, opened 2021-09-30
135 make secrets and their associated groups clearer

### LINKED ISSUE #135: Make Secrets and Their Associated Groups Clearer
Currently, a secret in group `dev` with name `api-password` will be given a k8s resource name of `dev-api-password`, which makes it hard to figure-out what group it belongs to, if you've forgotten what groups are in existence and you have other k8s secrets floating around. Maybe, this would be easier if the secret name would be something like `dev--api-password`?

Additionally, it would be useful to have a command along the lines of `bodywork secret display groups`, that lists the secrets groups in existence by looking at the labels of all secrets.

Marios85:
This PR closes #135 and contains changes purely related to formatting the `secrets display command`.

### DIFF
--- a/src/bodywork/cli/secrets.py
+++ b/src/bodywork/cli/secrets.py
@@ -38,7 +38,7 @@ def _parse_secret_key_value_pair(kv_string: str) -> Tuple[str, str]:
     key = kv_string[:equals_sign]
     if len(key) == 0:
         raise ValueError(error_msg)
-    value = kv_string[(equals_sign + 1) :]
+    value = kv_string[(equals_sign + 1):]
     if len(value) == 0:
         raise ValueError(error_msg)
     return key, value
@@ -142,10 +142,12 @@ def display_secrets(
         return None
     if secret_name:
         try:
-            complete_secret_name = _create_complete_secret_name(group, secret_name)
+            complete_secret_name = _create_complete_secret_name(
+                str(group), str(secret_name)
+            )
             print_dict(
                 secrets[complete_secret_name].data,
-                _create_table_name(secret_name, group),
+                _create_table_name(str(secret_name), str(group)),
             )
         except KeyError:
             print_warn(f"Cannot find secret={secret_name}.")
--- a/src/bodywork/cli/secrets.py
+++ b/src/bodywork/cli/secrets.py
@@ -38,7 +38,7 @@ def _parse_secret_key_value_pair(kv_string: str) -> Tuple[str, str]:
     key = kv_string[:equals_sign]
     if len(key) == 0:
         raise ValueError(error_msg)
-    value = kv_string[(equals_sign + 1):]
+    value = kv_string[(equals_sign + 1) :]
     if len(value) == 0:
         raise ValueError(error_msg)
     return key, value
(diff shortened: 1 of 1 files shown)


src/bodywork/cli/secrets.py
```diff
@@ -140,12 +140,21 @@ def display_secrets(
     if secret_name and not group:
         print_warn("Please specify which secrets group the secret belongs to.")
         return None
-    elif secret_name and group:
+    if secret_name:
```
AlexIoannides, line 143: I believe that `elif secret_name and group` means you don't need to cast sh*t as `str` later on, as your friend Mypy understand my version (which is why it is, the way it is) 🙂

src/bodywork/cli/secrets.py
```diff
@@ -140,12 +140,21 @@ def display_secrets(
...
         try:
-            complete_secret_name = _create_complete_secret_name(group, secret_name)
-            print_dict(secrets[complete_secret_name].data, complete_secret_name)
+            complete_secret_name = _create_complete_secret_name(
+                str(group), str(secret_name)
+            )
+            print_dict(
+                secrets[complete_secret_name].data,
+                _create_table_name(str(secret_name), str(group)),
```
AlexIoannides, line 150: Really shouldn't be repeat casting type just to appease mypy. I know it's frustrating, but...

src/bodywork/cli/secrets.py
```diff
@@ -140,12 +140,21 @@ def display_secrets(
...
+            )
         except KeyError:
             print_warn(f"Cannot find secret={secret_name}.")
+    elif group:
+        for secret in secrets.items():
+            table_name = _create_table_name(secret[0].split("-")[1], group)
```
AlexIoannides, line 156: I think it would be better to have,

```python
secret_name = secret[0].split("-")[1]
table_name = _create_table_name(secret_name, group)
```

It's much easier to read.

src/bodywork/cli/secrets.py
Marios85: Mypy forcing us to right less optimal code, i'm not cool with this. That means we need to do make a nested `if` which isn't as nice.

src/bodywork/cli/secrets.py
Marios85: Completely agree, can we disable MyPy for this sh*t because all i'm getting out of it, is being forced to write sub-optimal code due to MyPy's shortcomings.

src/bodywork/cli/secrets.py
AlexIoannides: I meant use,

`if secret_name and group:`

Instead of,

`if secret_name:`

I suggest we come up with a 'Mypy strategy' after v3. I am in agreement now - as it is, it's annoying.

src/bodywork/cli/secrets.py
AlexIoannides: The above change from the previous comment should fix this, as it's explicit enough for mypy to figure-out what's going wrong.

As I said before, let's leave this for now and come-up with a Mypy strategy (or replacement) after v3 is done. It may be, that we end-up configuring it just for function signatures or something.

src/bodywork/cli/secrets.py
Marios85: Sorry I should've explained with example above. This means we have an extra check for `group` because the following statement is `elif group` however it's probably better to refactor with the nested `if` to remove checking `group` three times:
```
    if secret_name:
        if not group:
            print_warn("Please specify which secrets group the secret belongs to.")
            return None
        else:
            try:
                complete_secret_name = _create_complete_secret_name(
                    str(group), str(secret_name)
                )
                print_dict(
                    secrets[complete_secret_name].data,
                    _create_table_name(secret_name, group),
                )
            except KeyError:
                print_warn(f"Cannot find secret={secret_name}.")
    elif group:
```

Either way Mypy has forced us into writing code that isn't as nice. My vote is still to disable Mypy for this but the above is the best compromise if we have to conform to Mypy.

PR #154, opened 2021-10-01
153 - add support for Jupyter notebooks

### LINKED ISSUE #153: Add support for Jupyter notebooks
"_As a ML Engineer, I would like to be able to define pipelines in Jupyter notebooks, so that I don't have to convert my notebooks into Python modules before I can deploy projects._"

**Tasks**

- add `nbconvert` and `nbformat` packages to Bodywork requirements.
- in `bodywork.stage_execution.run_stage`, check the file type of the configured executable and if it's `ipynb` then run the notebook in a different path, as demonstrated [here](https://github.com/bodywork-ml/bodywork-jupyter-pipeline-project/blob/master/stage_1_train_model/nb_runner.py).
- check config validation to make sure that `ipynb` files will pass validation for executable files.
- add one unit test and one integration test for this use-case (will need new test resources).

AlexIoannides:
This PR add support for using Jupyter notebooks to define stages and closes #153


src/bodywork/stage_execution.py
```diff
@@ -57,12 +68,23 @@ def run_stage(
...
+                notebook,
+                {"metadata": {"path": stage.executable_module_path.parent}},
+            )
+        else:
+            run(
+                ["python", stage.executable_module, *stage.args],
+                check=True,
+                cwd=stage.executable_module_path.parent,
+                encoding="utf-8",
+            )
```
Marios85, line 87: We need to think if we are going to add to the list of executables files that we will have in the future. If we do then the OO approach of a class with interface containing an `execute` method for each type would be nicer and extensible (open/closed principle).

tests/unit_and_functional/test_stage_execution.py
```diff
@@ -149,3 +154,30 @@ def test_run_stage_failure_raises_exception_for_failed_setup(
...
+def test_run_stage_with_jupyter_notebook(
+    setup_bodywork_test_project: Iterable[bool],
+    project_repo_connection_string: str,
+    bodywork_output_dir: Path,
+):
+    try:
+        run_stage("stage_jupyter", project_repo_connection_string)
+        assert True
+    except Exception:
+        assert False
```
Marios85, line 176: Do we need to print the exception out here so we can see the exception in the test runner (CircleCI)?

Marios85 (commented):
All fine, just the one comment really about maybe using an OO approach to executing files.

PR #155, opened 2021-10-07
151 simplify stage naming convention

### LINKED ISSUE #151: Simplify stage naming convention
**Background**
Currently, stages are named using the following convention:

```
PROJECT_NAME--STAGE_NAME
```

This can lead to very long service URLs. Given that the namespace is now nearly always `PROJECT_NAME`, we should simplify the stage name to be just `STAGE_NAME`, so that a service URL becomes something along the lines of,

```
http://CLUSTER_IP/weather-prediction/service/.../...
```

This may require a modification to the chosen fields in `bw deployment display` to make it obvious from which project the deployments come from.

Marios85:
This closes #151 . Stage name no longer contains the project-name. This PR also contains a bug fix for the git-commit hash, which may or may not have been added under another PR.

### DIFF
--- a/src/bodywork/k8s/workflow_jobs.py
+++ b/src/bodywork/k8s/workflow_jobs.py
@@ -41,7 +41,7 @@ def configure_workflow_job(
     project_repo_branch: str = "master",
     retries: int = 2,
     image: str = BODYWORK_DOCKER_IMAGE,
-    job_name: str = None,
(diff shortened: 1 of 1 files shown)


AlexIoannides:
@Marios85 I've changed so much (this was non-trivial), that I now need you to review what I've done...

PR #162, opened 2022-01-16
128 delete a group of secrets

### LINKED ISSUE #128: Delete a Group of Secrets
"As an ML Engineer I would like to delete a whole group of secrets"

Expand the CLI `delete secret` function so that it is possible to delete a whole group of secrets by just providing the group name. This will involve amending the existing 'delete secret' methods in both `k8s.secrets.py` & `cli.secrets.py`.

Marios85:
This PR closes #128 . 

The existing secret command is used to delete a whole group of secrets by entering the following command `bodywork delete secret --group test`.


src/bodywork/cli/secrets.py
```diff
@@ -123,6 +123,22 @@ def delete_secret(namespace: str, group: str, secret_name: str) -> None:
...
+    """Delete a group of secrets within a k8s namespace.
+
+    :param namespace: Namespace in which to look for secrets group.
+    :param group: The group of secrets to delete.
+    """
+    if not k8s.namespace_exists(namespace):
+        print_warn(f"Could not find namespace={namespace} on k8s cluster.")
+        return None
+    if not k8s.secret_group_exists(namespace, group):
+        print_warn(f"Could not find secret group={group} .")
```
AlexIoannides, line 136: Don't need a blank space before the full-stop.

src/bodywork/cli/secrets.py
```diff
@@ -123,6 +123,22 @@ def delete_secret(namespace: str, group: str, secret_name: str) -> None:
...
+        return None
+    k8s.delete_secret_group(namespace, group)
+    print_info(f"Deleted secret group={group} .")
```
AlexIoannides, line 139: Don't need a blank space before the full-stop.

src/bodywork/k8s/secrets.py
```diff
@@ -143,6 +143,25 @@ def secret_exists(
         return False
 
 
+def secret_group_exists(namespace: str, group: str) -> bool:
+    """ Does the specified secret group exist
```
AlexIoannides, line 147: Doctoring format needs adjusting so that there isn't a blank space before the start of the first line (needs a full-stop, too). And then the items below shouldn't be indented.

src/bodywork/k8s/secrets.py
```diff
@@ -143,6 +143,25 @@ def secret_exists(
...
+
+        :param namespace: Kubernetes namespace in which to look for secrets group.
+        :param group: Name of secrets group.
+        :return: True if group exists, otherwise False.
+    """
+    return (
+        len(
```
AlexIoannides, line 154: Do you mind unpacking this?

src/bodywork/k8s/secrets.py
```diff
@@ -193,12 +212,23 @@ def delete_secret(namespace: str, name: str) -> None:
     k8s.CoreV1Api().delete_namespaced_secret(namespace=namespace, name=name)
 
 
+def delete_secret_group(namespace: str, group: str) -> None:
+    """Delete a group of secrets in a namespace.
+
+    :param namespace: Namespace in which to look for the secret to
+    delete.
```
AlexIoannides, line 219: Should be indented.

src/bodywork/k8s/secrets.py
```diff
@@ -193,12 +212,23 @@ def delete_secret(namespace: str, name: str) -> None:
...
+    :param group: The name of the secrets group to be deleted.
+    """
+    k8s.CoreV1Api().delete_collection_namespaced_secret(
+        namespace=namespace, label_selector=f"{SECRET_GROUP_LABEL}={group}"
```
AlexIoannides, line 223: Nice!

AlexIoannides (approved):
I have some trivial comments that I'd appreciate if you take a look at, but otherwise it's a 1st class job (as always).

src/bodywork/k8s/secrets.py
Marios85: I agree it's horrible formatting but this is Black, so it's going to do it every time we run it.

src/bodywork/k8s/secrets.py
Marios85: Good spot, my IDE used to generate these automatically but it's not doing it after I got my laptop repaired for some reason.

src/bodywork/k8s/secrets.py
Marios85: Should all be on the same line, fixed this and the comment for the delete secrets method as well.

src/bodywork/k8s/secrets.py
Marios85: 🙂

src/bodywork/k8s/secrets.py
Marios85: It's a little bit better with the conditional on another line.

PR #165, opened 2022-02-22
Make Git branch an optional argument

AlexIoannides:
This PR makes the specification of a Git repo's branch an optional parameter (when creating deployments, etc.). If it is unspecified, then it will use the 'default' branch for that repository (e.g., 'master' or 'main', or whatever has been setup behind-the-scenes). 

The motivation behind this, is to make the CLI even cleaner. If the branch needs to be specified, then the user can use the `--branch` flag.

As part of the PR, I also performed the following maintenance tasks:

- I updated the dates on the copyright notice to include 2022;
- I upgraded all of our development dependencies (including mypy...); and,
- I removed `jupyter` as a package dependency and replaced it with `ipykernel` (I don't think we can get away without this).


Marios85 (approved):
Everything looks good! Nice that it can default to whatever the master/main branch is called, I didn't know we could do that.

PR #166, opened 2022-02-22
Rename 'project' to 'pipeline' in bodywork.yaml spec

AlexIoannides:
This PR renames 'project' to 'pipeline' in bodywork.yaml spec, to bring it in-line with the terminology used throughout version three.

This PR will require `rc3` to be merged first (so that I can then update the various test projects to use the new config schema), but I have tested this manually on the `v3` branch of `bodywork-test-project` and it looks good.


Marios85 (approved):
Nothing to say, all fine 😎

PR #167, opened 2022-02-27
Pre-release Housekeeping

AlexIoannides:
This PR takes care of the following tasks prior to releasing version 3:

- Update README.
- Hide all undocumented CLI options from CLI help.
- Change cronjob history table name from `workflow job` to `run ID`.
- Upgrade Kubernetes Python client.


src/bodywork/cli/cli.py
```diff
@@ -180,11 +180,11 @@ def _debug(seconds: int = Argument(600)) -> None:
...
     git_url: str = Argument(...),
     git_branch: str = Option("", "--branch"),
-    asynchronous: bool = Option(False, "--async"),
-    asynchronous_job_name: str = Option("", "--async-job-name"),
+    asynchronous: bool = Option(False, "--async", hidden=True),
+    asynchronous_job_name: str = Option("", "--async-job-name", hidden=True),
     ssh_key_path: str = Option("", "--ssh"),
     secrets_group: str = Option("", "--group", "--secrets-group"),
-    image: str = Option(None, "--bodywork-image"),
+    image: str = Option(None, "--bodywork-image", hidden=True),
```
Marios85, line 187: I don't think we should hide the option to use a custom image. This is a useful feature and it's straightforward so I don't think it will confuse anyone. We also mention that you can override the image in the docs so we should keep the option there for people to see.

README.md
```diff
@@ -13,32 +13,59 @@
...
+
+Stages are assembled into [DAGs](https://en.wikipedia.org/wiki/Directed_acyclic_graph) that define your pipeline's workflow. This and other key [configuration](user_guide.md#configuring-a-project-for-deployment-with-bodywork) is contained within a single [bodywork.yaml file](https://github.com/bodywork-ml/bodywork-ml-pipeline-project/blob/master/bodywork.yaml).
+
+## Simplified DevOps for ML
+
+Bodywork removes the need for you to build and manage container images for any stage of your pipeline. It works by running all stages using Bodywork's [custom container image](https://hub.docker.com/repository/docker/bodyworkml/bodywork-core). Each stage starts by pulling all the required files directly from your project's Git repository (e.g., from GitHub), pip-installing any required dependencies, and then running the stage's designated Python module (or Jupyter notebook).
+
+<div align="center">
+<img src="https://bodywork-media.s3.eu-west-2.amazonaws.com/ml_pipeline.png"/>
+</div>
```
Marios85, line 49: I think we should add a line in somewhere just saying that it works on any K8s platform, no matter where it's hosted to be clear. This is one of our important features and likely something someone will want to know from the off.

Marios85 (commented):
All looks good, just a couple of in-line comments.

PR #171, opened 2022-04-23
Tech debt - skip git SSH tests when running locally

AlexIoannides:
It can be difficult to setup SSH keys to run the integration tests for SSH based connectivity with our private repos on GitHub, GitLab, BitBucket and Azure DevOps. To make local development easier, we will skip these tests when not running on CICD.


AlexIoannides:
@Marios85 - I'll let you finish your work, which includes the required updates of Typer, before actually merge this one.

PR #172, opened 2022-04-26
Bug fix - unintended loading of kubeconfig file

AlexIoannides:
This PR fixes a bug with the `bodywork.cli.cli.k8s_auth` function, that was causing the `load_kubernetes_config` function to be called every time a function was wrapped by `k8s_auth`, and not when the wrapped function was itself called (the intended behaviour).


Marios85 (approved):
🙂

Issue #169, opened 2022-04-21
bodywork get deployment - GitBranch Shows Stage Name Instead of Git Branch. 


Marios85:
When viewing a deployment using `bodywork get deployment` the stage name is incorrectly shown for GitBranch. 

See Attached email for detail.
[Email - Marios Ioannides - Outlook.pdf](https://github.com/bodywork-ml/bodywork-core/files/8529769/Email.-.Marios.Ioannides.-.Outlook.pdf)

AlexIoannides:
Closed by #173

PR #173, opened 2022-04-27
#169 bodywork get deployment   git branch shows stage name instead of git branch

Marios85:
Fixed so that the git-branch is shown and removed the stage name field as this is the same as the `name` field, therefore displaying duplicate data.

This commit also contains a fix for the CI/CD build:

- Upgrade deprecated Ubuntu image for docker builds.
- Fix Docker Dev Build Step after upgrade.


AlexIoannides (approved):

AlexIoannides:
BTW - I think the version should be changed to `3.0.1` - there's no need for the `dev` tag as this is bona fide patch.

Marios85:
Cheers, i'll change the version and then merge.

PR #177, opened 2022-05-01
Improve stage logging

AlexIoannides:
This PR introduces some extra log messages to stage execution, and removes the `capture_logging=True` argument from the subprocess that installs requirements, so that everything is printed to stdout.

Test coverage has also been improved, so that tests will fail if any of this additional information is not logged or printed to stdout, as required.


Marios85 (approved):
So `capture_output=True` stops it printing to the console and consequently, to the logs. 

Nice1 for fixing this 😎

PR #178, opened 2022-05-12
Improve cluster resource monitoring and logging

AlexIoannides:
This PR introduces a new function, `bodywork.k8s.utils.has_unschedulable_pods`, that checks if any of the pods associated with a job or deployment cannot be scheduled by the cluster - i.e., if there are inadequate resources (cpu and/or memory). 

This k8s notification is raised very quickly, so that when this new function is called prior to job or deployment monitoring, it can be used to shut down a deployment without having to wait unnecessarily for minutes and to then report this event specifically - e.g.,

```
$ bw create deployment https://github.com/bodywork-ml/bodywork-serve-model-project.git  --branch test  

Using local workflow controller - retries inactive.
========================================= deploying test branch from https://github.com/bodywork-ml/bodywork-serve-model-project.git ==========================================
[05/12/22 15:04:49] INFO     Creating k8s namespace = bodywork-serve-model-project                                                                                             
[05/12/22 15:04:49] INFO     Creating k8s service account = bodywork-stage                                                                                                     
[05/12/22 15:04:49] INFO     Attempting to execute DAG step = [scoring_service]                                                                                                
[05/12/22 15:04:49] INFO     Creating k8s deployment and service for stage = scoring-service                                                                                   
[05/12/22 15:04:54] ERROR    Deployment failed --> Inadequate cluster cpu and/or memory available to start deployments (scoring-service)                                       
[05/12/22 15:04:55] INFO     Searching for services from previous deployment.  
```

Previously, the user could wait for minutes for the timeout to be raised and would be met with empty pod logs - i.e., none the wiser about why the deployment failed.


src/bodywork/workflow_execution.py
```diff
@@ -303,14 +303,15 @@ def _run_batch_stages(
...
+        _print_logs_to_stdout(namespace, job_name)
     except TimeoutError as e:
         _log.error("Some (or all) k8s jobs failed to complete successfully")
+        _print_logs_to_stdout(namespace, job_name)
         raise e
     finally:
         for job_object in job_objects:
             job_name = job_object.metadata.name
             _log.info(f"Completed k8s job for stage = {job_name}")
-            _print_logs_to_stdout(namespace, job_name)
```
Marios85, line 313: I don't understand this change. When `_print_logs_to_stdout(namespace, job_name)` is in the `finally` block then all logs should be printed out for each job when they've completed or timed out. 

Now the only logs that will be printed are for the last job in the loop defined on line 300. Is this correct or have I missed something?

src/bodywork/k8s/deployments.py
```diff
@@ -333,6 +334,15 @@ def monitor_deployments_to_completion(
...
     sleep(wait_before_start_seconds)
+
+    unschedulable_pods = [
+        has_unscheduleable_pods(deployment) for deployment in deployments
+    ]
+    if any(unschedulable_pods):
+        raise BodyworkClusterResourcesError(
+            "deployment", [deployment.metadata.name for deployment in deployments]
+        )
+
```
Marios85, line 345: Could also add this as a method to `utils.py` (e.g. check_cluster_resources) because it's in batch_jobs.py as well. It's only a tiny bit of repeated code but it will make it more readable as the formatting splits it on multiple lines.

Marios85 (commented):
Happy that you've fixed this so Bodywork actively monitors for resource failures. 

A couple of inline comments but otherwise it looks great, as per usual 🙂

src/bodywork/workflow_execution.py
AlexIoannides: If there are no resources available to run a pod, then it won't generate any logs. So, I don't want to print empty pod logs to stdout. Thus, the logs will only now get printed if the job/deployment complete successfully or times-out.

AlexIoannides: But you're right, I am missing a for loop...

src/bodywork/k8s/deployments.py
AlexIoannides: I think I didn't do this, because of the need to pass the resource type to the exception. I'll have another think about this, because I agree with the suggestion.

src/bodywork/workflow_execution.py
Marios85: Cool, that makes sense. `k8s.monitor_jobs_to_completion` can also throw up a `BodyworkJobFailure` exception so I think we need to catch that exception and print the logs out for it as well (e.g. in the same block as TimeoutError).

src/bodywork/k8s/deployments.py
AlexIoannides: I've sorted this out - much better 👍

Marios85 (approved):
All good! 🙂

PR #179, opened 2022-05-21
Make workflow orchestration more robust and predictable

AlexIoannides:
This PR covers a collection of small enhancements that aim to make workflow orchestration more robust and predictable - i.e., make Bodywork feel less "brittle".

## Where useful, get pod logs from crashed containers

Enhanced `bodywork.k8s.pod_logs.get_pod_logs` and `bodywork.workflow_execution._print_logs_to_stdout` (that uses the latter), so that the workflow can choose to retrieve pod logs from a previously crashed container, in their entirety. This guarantees no missing logs when getting logs for stages that have been killed by a timeout expiring.

## Use startup and liveness probes to accurately determine the state of a service stage

Added startup and liveness probes to deployments, so that the mechanism for detecting that a service stage is 'ready', is dependent on the server being up (it does a TCP ping), as opposed to relying on the container not crashing for `max_startup_time_seconds` (taken from `bodywork.yaml`). The `max_startup_time_seconds` config parameter is now used to define how long to wait for requirements to install and the server to start. This makes integration tests *much* more predicable 🙂

## Enable precise timeouts for batch stages

In `bodywork.stage_execution.run_stage` there is a new `timeout` argument, that is used for running batch stages. This enables us to the `max_completion_time_seconds` config parameter to refer solely to the time required to run the job, as opposed to lumping it together with installing cloning the repo, installing requirements, etc. 

## Tune the rolling update strategy for deployments

Perviously, we were using the default parameters, but now we create one new pod and delete an existing deployment's pod, to start creating the pods for the incoming deployment. This also enables us to more accurately compute the required resources for the update, together with how long an updated deployment might take to roll-out...

## Stage monitoring timeouts are now easier to understand

The stage monitoring timeouts in `bodywork.workflow_execution` have been modified in light of the above changes, and are now calculated in dedicated functions (one for jobs and one for deployments). In particular, a deployment is now allowed to restart at most two time before halting the rollout (with some buffer for pulling images and backoff between restarts). The batch-job timeout is now purely a 'just in case something unexpected happens', because between retries and timeout monitoring (inside container), this will now get handled automatically.

## Wait for namespaces to be deleted

Small update to `bodywork.k8s.namepaces.delete_namespace` so that it now waits for the namespace to stop existing, before returning. This prevents users from deleting a deployment and re-creating it before the namespace has finished deleting, which results in an error. This also helps with our integration tests...


src/bodywork/k8s/deployments.py
```diff
@@ -75,18 +81,21 @@ def configure_service_stage_deployment(
...
-        [project_repo_url, stage_name, f"--branch={project_repo_branch}"]
-        if project_repo_branch
-        else [project_repo_url, stage_name]
-    )
+
+    container_args = [project_repo_url, stage_name]
+    if project_repo_branch:
+        container_args += [f"--branch={project_repo_branch}"]
+
+    probe_period_seconds = 10
```
Marios85, line 96: Add this to `constants.py`

src/bodywork/stage_execution.py
```diff
@@ -89,11 +93,16 @@ def run_stage(
...
                 cwd=stage.executable_module_path.parent,
                 encoding="utf-8",
+                timeout=timeout,
             )
         _log.info(
             f"Successfully ran stage = {stage_name} from {repo_branch} branch of repo "
             f"at {repo_url}"
         )
+    except TimeoutExpired:
+        msg = f"Timout exceeded when running {stage.executable_module}"
```
Marios85, line 103: Typo

src/bodywork/workflow_execution.py
```diff
@@ -301,16 +306,25 @@ def _run_batch_stages(
...
+        timeout_dt = (datetime.now() + timedelta(seconds=timeout)).strftime(
+            "%d/%m/%y %H:%M:%S"
+        )
+        _log.info(f"Monitoring k8s jobs, timeout at [{timeout_dt}] ({timeout}s)")
+        k8s.monitor_jobs_to_completion(job_objects, timeout)
+        for job_object in job_objects:
+            job_name = job_object.metadata.name
+            _log.info(f"Completed k8s job for stage = {job_name}")
+            _print_logs_to_stdout(namespace, job_name, False)
+    except Exception as e:
```
Marios85, line 319: Shouldn't this be `(TimeoutError, BodyworkJobFailure)` to catch everything from `monitor_jobs_to_completion()`? I'm also cool with a general Exception in this case, if there is a possibility that a legitimate unexpected Exception could be thrown by `monitor_jobs_to_completion()` with retrievable pod logs. 

Otherwise there should be a different handling for general exceptions.

src/bodywork/workflow_execution.py
```diff
@@ -369,15 +383,17 @@ def _run_service_stages(
...
             k8s.create_deployment(deployment_object)
     try:
-        timeout = max(stage.max_startup_time for stage in service_stages)
-        k8s.monitor_deployments_to_completion(
-            deployment_objects, timeout + TIMEOUT_GRACE_SECONDS
+        timeout = _compute_optimal_deployment_timeout(namespace, service_stages)
+        timeout_dt = (datetime.now() + timedelta(seconds=timeout)).strftime(
+            "%d/%m/%y %H:%M:%S"
         )
+        _log.info(f"Monitoring k8s deployments, timeout at [{timeout_dt}] ({timeout}s)")
```
Marios85, line 390: Giving the ETA to the user is a really good idea!

src/bodywork/workflow_execution.py
```diff
@@ -301,16 +306,25 @@ def _run_batch_stages(
...
         k8s.create_job(job_object)
     try:
-        timeout = max(stage.max_completion_time for stage in batch_stages)
-        k8s.monitor_jobs_to_completion(job_objects, timeout + TIMEOUT_GRACE_SECONDS)
-    except TimeoutError as e:
+        timeout = _compute_optimal_job_timeout(batch_stages)
+        timeout_dt = (datetime.now() + timedelta(seconds=timeout)).strftime(
+            "%d/%m/%y %H:%M:%S"
+        )
+        _log.info(f"Monitoring k8s jobs, timeout at [{timeout_dt}] ({timeout}s)")
```
Marios85, line 313: Giving the ETA to the user is a really good idea!

src/bodywork/workflow_execution.py
```diff
@@ -490,16 +506,63 @@ def parse_dockerhub_image_string(image_string: str) -> Tuple[str, str]:
...
+    :param batch_stages: The desired configuration for the jobs.
+    :param returns: The optimal timeout (in seconds).
+    """
+    job_timeouts = [
+        max(1, stage.retries) * stage.max_completion_time for stage in batch_stages
+    ]
+    return int(max(job_timeouts) + TIMEOUT_GRACE_SECONDS)
+
+
+def _compute_optimal_deployment_timeout(
```
Marios85, line 522: I think these methods should be moved to `batch_jobs.py` and `deployments.py` respectively, as they refer to those specific things. Also `workflow_execution.py` is a pretty big file as it is.

What do you think?

src/bodywork/workflow_execution.py
```diff
@@ -490,16 +506,63 @@ def parse_dockerhub_image_string(image_string: str) -> Tuple[str, str]:
...
+    namespace: str, service_stages: List[ServiceStageConfig]
+) -> int:
+    """Compute the optimal timeout for deployment monitoring.
+
+    If a deployment is rolling-out for the first time, then wait for
+    twice the max configured startup time. If a deployment is updating,
+    then compute how long it will take to replace all pods using the
+    rolling update strategy.
+
+    The the max configured startup time has been floored at 60s, because
```
Marios85, line 532: Typo

src/bodywork/workflow_execution.py
```diff
@@ -490,16 +506,63 @@ def parse_dockerhub_image_string(image_string: str) -> Tuple[str, str]:
...
+    new_pod_rate = K8S_MAX_SURGE + K8S_MAX_UNAVAILABLE
+    deployment_timeouts = [
+        (
+            stage.max_startup_time
+            if not k8s.is_existing_deployment(namespace, stage.name)
+            else ceil(stage.replicas / new_pod_rate) * max(60, stage.max_startup_time)
+        )
+        for stage in service_stages
+    ]
+    return int(2 * max(deployment_timeouts) + TIMEOUT_GRACE_SECONDS)
```
Marios85, line 550: Might be a good idea to add a one liner in the docs somewhere to explain about the min startup time etc.

src/bodywork/workflow_execution.py
```diff
@@ -490,16 +506,63 @@ def parse_dockerhub_image_string(image_string: str) -> Tuple[str, str]:
...
+    If a deployment is rolling-out for the first time, then wait for
+    twice the max configured startup time. If a deployment is updating,
+    then compute how long it will take to replace all pods using the
+    rolling update strategy.
+
+    The the max configured startup time has been floored at 60s, because
+    installing just Flask alone takes this long and it would be easy to
+    incorrectly estimate this.
+
+    :param namesapce: The target namespace for the deployments.
```
Marios85, line 536: Just spotted another typo...i'm being mega anal today 🙂

Marios85 (changes requested):
Yes, yes and yes! So much good stuff in this PR! 🙂 This guarantees feedback to the user and makes Bodywork more "predictable", which I agree, will make Bodywork feel a lot less "brittle". 

A few relatively minor comments inline and that's it.

src/bodywork/k8s/deployments.py
AlexIoannides: Good spot.

src/bodywork/stage_execution.py
AlexIoannides: Hawkeye!

src/bodywork/workflow_execution.py
AlexIoannides: I agree with your suggestions.

`monitor_jobs_to_completion` could throw a k8s API exception, but this isn't where we should be handling it (we should do that within the function itself (job for another day).

src/bodywork/workflow_execution.py
AlexIoannides: I thought it was a good idea at face value, but when I copied `_compute_optimal_job_timeout` into `bodywork.k8s.batch_jobs`, I realised that `BatchStageConfig` would need to be imported and that this isn't really a k8s entity - it's a Bodywork entity.

So I'm going to leave this out for the time being.

src/bodywork/workflow_execution.py
AlexIoannides: Agreed. I've made a note of this for when it's released.

src/bodywork/workflow_execution.py
AlexIoannides: Personally, I really like this 🙂

src/bodywork/stage_execution.py
```diff
@@ -89,11 +93,16 @@ def run_stage(
...
+        stage_failure_exception = BodyworkStageFailure(stage_name, msg)
+        raise stage_failure_exception
```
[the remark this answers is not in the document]
AlexIoannides: Agreed.

Marios85 (approved):
Looks good!

src/bodywork/workflow_execution.py
Marios85: That's a good point, it is more of a config entity.

PR #180, opened 2022-05-23
Add progress bar to job and deployment monitoring

AlexIoannides:
This PR adds a (transient) progress bar to k8s job and deployment monitoring, as shown below.

image

It updates the current time and progress to timeout while waiting for k8s, to demonstrate to the user that they need to wait and that Bodywork has not hung. The aim is to tackle the perception of "brittleness" that comes for staring at a static screen. When the workflow progresses the progress bar is removed - i.e., it will not be retained in logs.

In the future, this can be extended to add more information - e.g., number of pods replaced during a deployment update, etc. But, we have more-than-enough for now.


AlexIoannides:
Refactored as asked.

Marios85 (approved):
Perfect! Thanks for doing that!

PR #183, opened 2022-06-22
Handle keyboard interrupts during workflow execution

AlexIoannides:
This PR adds `KeyboardInterrupt` to the set of exceptions that are handled during batch and service stage execution, alongside `TimeoutError`. Thus, if a user gets tired of waiting (e.g., they are just testing), then they can terminate the workflow gracefully.


Marios85 (approved):
😎

PR #194, opened 2022-07-05
Change relative to absolute imports

gadc1996:


AlexIoannides:
Looks great!

If you could bump the version in the `VERSION` file to `3.0.13`, then the tests will run on our CICD server and we can work towards getting this merged.

If you could also run [black](https://black.readthedocs.io/en/stable/) on the codebase, then that will fix any flake8 that might have crept in.

We'll post all this info somewhere sensible in the coming days.

AlexIoannides:
PR has gone stale.

